Issue #9, opened 2016-02-10
Visual Studio project generation not including appropriate headers


Nuclearfossil:
After generating a Visual Studio solution using the following:

```
cmake.exe -G "Visual Studio 14 2015 Win64"
```

Visual studio fails in compilation with the following:

```
1>------ Build started: Project: EASTL, Configuration: Debug x64 ------
1>  allocator_eastl.cpp
1>C:\Code\Research\EASTL\include\EASTL/internal/config.h(61): fatal error C1083: Cannot open include file: 'EABase/eabase.h': No such file or directory
1>  assert.cpp
1>C:\Code\Research\EASTL\include\EASTL/internal/config.h(61): fatal error C1083: Cannot open include file: 'EABase/eabase.h': No such file or directory
1>  fixed_pool.cpp
1>C:\Code\Research\EASTL\include\EASTL/internal/fixed_pool.h(22): fatal error C1083: Cannot open include file: 'EABase/eabase.h': No such file or directory
1>  hashtable.cpp
1>C:\Code\Research\EASTL\include\EASTL/internal/hashtable.h(30): fatal error C1083: Cannot open include file: 'EABase/eabase.h': No such file or directory
1>  intrusive_list.cpp
1>C:\Code\Research\EASTL\include\EASTL/internal/config.h(61): fatal error C1083: Cannot open include file: 'EABase/eabase.h': No such file or directory
1>  numeric_limits.cpp
1>C:\Code\Research\EASTL\include\EASTL/internal/config.h(61): fatal error C1083: Cannot open include file: 'EABase/eabase.h': No such file or directory
1>  red_black_tree.cpp
1>C:\Code\Research\EASTL\include\EASTL/internal/config.h(61): fatal error C1083: Cannot open include file: 'EABase/eabase.h': No such file or directory
1>  string.cpp
1>C:\Code\Research\EASTL\include\EASTL/internal/config.h(61): fatal error C1083: Cannot open include file: 'EABase/eabase.h': No such file or directory
1>  thread_support.cpp
1>C:\Code\Research\EASTL\include\EASTL/internal/config.h(61): fatal error C1083: Cannot open include file: 'EABase/eabase.h': No such file or directory
1>  Generating Code...
2>------ Skipped Build: Project: ALL_BUILD, Configuration: Debug x64 ------
2>Project not selected to build for this solution configuration 
```

Nuclearfossil:
Just submitted a PR to resolve this.

[...]

PR #10, opened 2016-02-10
Potential fix for https://github.com/electronicarts/EASTL/issues/9

Nuclearfossil:
This is a PR fix for:

1. Fixing cmakelist.txt to look in the test/packages/EABase/include/Common for additional required headers.
2. Updating .gitignore it ignore VS 2015 build files.


rparolin:
Looks good other than the cmake version changes I pointed out.   Thanks!

Nuclearfossil:
So, as an aside, where is it documented that cmake needs to be run from the test folder?  I'm actually more used to the process of:

1. create a 'build' folder.
2. call cmake from that build folder like:
``` cmake -g "whatever" ...

**Comment by rparolin** [comment]
Where is the test folder documented?  I have a section on how to build the unit tests in the CONTRIBUTING file at the root.

```
Here is an example batch file.
set build_folder=out
call cmake -B%build_folder% -Htest
pushd %build_folder%
call cmake --build . --config Release
call cmake --build . --config Debug
call cmake --build . --config RelWithDebInfo
call cmake --build . --config MinSizeRel
call ctest -C Release
call ctest -C Debug
call ctest -C RelWithDebInfo
call ctest -C MinSizeRel
popd
```

PR #16, opened 2016-02-13
A fix for compilation on Ubuntu

Draghi:
I couldn't compile the test on Ubuntu, so I attempted to fix the errors.

- Set the CMAKE_CXX_STANDARD flag to 11, which fixes the error(s) regarding auto.
- Changed the old-style cast in EAMain to a static_cast, this seemed to fix the error (so did removing the cast altogether)
- Included the EABase/eastdarg.h in EAStdC/EASprintf.h to remove the error regarding va-end.
- Included cwchar in EAStdC/EASprintf.cpp which seems to have gotten rid of the redefinition error.
- Changed initialisation for gGlobalErrorCount in EATest.cpp to a constructor with an cast integer, at least on my system the compiler was trying to call the deleted copy function due to the cast overload.
- Added a Linux implementation of SetThreadPriority, it's definitely not ideal but it seems to do it's job, comments about the issues are included in the implementation.
- Added a missing semi-colon in EASTLTestAllocator.cpp

This allows it to compile on my system and pass the test. I haven't tested it on windows since I'm away from my desktop computer at the moment. This pull request should also fix some compilation errors on mac, I believe, however mac will still need a SetThreadPriority implementation.


Draghi:
If I revert the change to the change to AtmoicInt my compiler spits out this error:
http://pastebin.com/m4JwVtaG

It appears that GCC is constructing an AtmoicInt out of the number and then trying to copy-construct gGlobalErrorCount despite the copy constructor being deleted. Simply adding a cast there doesn't relieve the error, however it does actually compile fine without the cast, but with an explicit constructor call.

------------

The issue actually appears to be with differing implementations between Visual Studio and gcc.

Visual studio defines a copy constructor for atmoic variables:
https://msdn.microsoft.com/en-us/library/hh874820.aspx

Where as gcc explicitly deletes the copy constructor for atomic variables which in turn means that the default copy-constructor for AtmoicInt gets deleted implicitly. This is with GCC version 4.8.4

If I add a copy constructor to AtmoicInt it also compiles. The issue being of course that the operation isn't atomic.

------------

I'll get right on fixing those comments up.

Draghi:
Fixed comment style.

rparolin:
Ah.  Nice catch with the MS vs stdlibc++ difference.  EAThread Atomic<T> are copyable internally as well so lets implement the copy constructor.  It's not perfect but EAThread Atomic<T> is really intended for integral types not arbitrary user types.

Unfortunately Travis-CI is still reporting errors after your changes.  It definite improvement the gcc error its reporting surprises me if you built cleanly.

Other than that looks good.  Run a spell checker over your thread priority comments though I noticed some spelling mistakes.

Thanks!

rparolin:
Ah.  Travis is using GCC 4.6.  Doh.

Draghi:
Apparently a fix for "sorry, unimplemented: mangling dotstar_expr" was only introduced in GCC 4.7 going by this report (even though the report isn't for this specific error):
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=48051

I've added the copy constructor and fixed the spelling mistakes (Hopefully I got them all)

rparolin:
Thanks!  Looks cool.  I'll look into getting travis-ci up on GCC 4.8 now.

PR #25, opened 2016-02-16
Updated the CMake build system to work with Intel's compilers on Unix based systems, including OS X.

breannansmith:
CMake does not support compiler traits with the Intel compiler (used in the root CMakeLists.txt file for 'set_property(TARGET EASTL PROPERTY CXX_STANDARD 11)'), and the Intel compiler requires an explicit asm-block flag to build. This pull request checks if the build is executing on a Unix-based system with the Intel compiler, and adds flags to enable C++11 and asm-blocks. This enables successful builds on OS X Mavericks and Ubuntu 12.04 with Intel 16.0.1.


rparolin:
Cool.  Thanks!

PR #29, opened 2016-02-21
Include header files in project sources

lukaszdk:
Allows you to browse the EASTL headers from the generateed VS/Xcode/etc. project.


rparolin:
Thank you

PR #30, opened 2016-02-22
Updated docs to Markdown

CrociDB:
Updated the doc files to Markdown, so it can be viewed directly on GitHub.


rparolin:
That's awesome.  This is something I've been thinking about doing.  Did you use a tool to convert the html to md?  I'm required to maintain the html documentation format internally.  If you used a tool, then I can keep the md files in sync with the html files without having to update the docs in two places.

CrociDB:
Hey!

I've thought about that too. I tried to convert HTML to MD using available online tools, but all of them would generate some garbage or some parts would not work properly on github. So I fixed all the problems and thought that generating HTML from the MD was actually the best choice. It's easier to edit in MD and there are a lot of tools that would do that. I thought of writing a simple tool in python to compile them to HTML myself, but I was afraid that wouldn't meet the project's guideline or something like that.

What do you think about it?

Thanks.

rparolin:
Unfortunately, I need to support the current HTML documentation format and I want one place to update documentation.  So if the conversion process isn't automatically then its going to be too cumbersome to use.  That being said, I think its a good idea to have a README.md file that links to the HTML docs using rawgit links.  That would (by some definition) make the links viewable in Github and I could include that easily.

Sorry.  I feel bad turning away your significant amount of manual work. Especially since its a popular request.  :(

rparolin:
Any interest in resurrecting this pull request?  I was thinking about do this work but I wanted to give you the opportunity to have this original PR make it into the repo.   Thanks.

PR #32, opened 2016-03-06
CMake restructuring

pzychotic:
I changed the CMake files so everything can be build from the top-level CMake, as probably most people would expect. This also removes some awkward references from the test and benchmark directories to its parent directory. And it also eliminates the use of undocumented CMake command line options to get the tests and benchmark work from its respective sub-directories.
I also put a workround in place to have all the source and header files show in IDE projects, that makes it a lot easier to work on the whole project and all its dependencies.

Tell me what you think of it or if this is going too far.


pzychotic:
Hmmm... runs without problems locally with VS2015, trying to figure out why it fails here.

pzychotic:
For some reason AppVeyor isn't using my changed build script, the command line still has the old CMake call with 'cmake -Bbuild -Htest'.
I compiled the changes locally on the command line with the build steps from the appveyor.yml script and it works.

.travis.yml
```diff
@@ -27,29 +27,25 @@ install:
...
 # Run tests 
-  - cd build
-  - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then ../cmake-3.4.3-Linux-x86_64/bin/ctest -C Release -V                                           ;fi
-  - if [[ "$TRAVIS_OS_NAME" == "osx"   ]]; then ../cmake-3.4.3-Darwin-x86_64/CMake.app/Contents/bin/ctest -C Release -V                       ;fi
+  - cd test
+  - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then ../../cmake-3.4.3-Linux-x86_64/bin/ctest -C Release -V                                           ;fi
+  - if [[ "$TRAVIS_OS_NAME" == "osx"   ]]; then ../../cmake-3.4.3-Darwin-x86_64/CMake.app/Contents/bin/ctest -C Release -V                       ;fi
 # Run benchmarks 
   - cd ..
-  - cd build_bench
```
rparolin: I'd prefer to keep all the binaries in a directory names that clearly implies this.  So I'd prefer something like:
* build/test
* build/bench

CMakeLists.txt
```diff
@@ -2,11 +2,27 @@
...
+# Sub-projects
+#-------------------------------------------------------------------------------------------
+add_subdirectory(test/packages/EAAssert)
+add_subdirectory(test/packages/EABase)
+add_subdirectory(test/packages/EAStdC)
+add_subdirectory(test/packages/EAMain)
+add_subdirectory(test/packages/EATest)
+add_subdirectory(test/packages/EAThread)
+
+add_subdirectory(benchmark)
```
rparolin: When using EASTL as a library, it shouldn't be referencing the tests or the benchmarks.

benchmark/CMakeLists.txt
```diff
@@ -6,7 +6,7 @@
 # CMake info 
 #-------------------------------------------------------------------------------------------
 cmake_minimum_required(VERSION 3.1)
-project(EASTLBenchmarks)
+project(EASTLBenchmarks CXX)
```
rparolin: Is there a benefit to explicitly including the 'CXX'?  The documentation I've read says that the project enabled C/CXX by default.  https://cmake.org/cmake/help/v3.0/command/project.html

test/packages/EAStdC/CMakeLists.txt
```diff
@@ -2,11 +2,25 @@
...
 
 #-------------------------------------------------------------------------------------------
 # Library definition 
 #-------------------------------------------------------------------------------------------
-file(GLOB EASTDC_SOURCES "source/*.cpp")
+set(HEADER_DIR "include/EAStdC")
+set(SOURCE_DIR "source")
+set(EASTDC_SOURCES
+    ${HEADER_DIR}/EAAlignment.h
+    ${HEADER_DIR}/EADateTime.h
```
rparolin: I prefer to have the GLOB style syntax so that I don't I have to edit build scripts when I add/remove files.  This also mimics the behavior of our internal build system.

rparolin:
I've added a bunch of comments to the PR.  Clearly the build error on Windows needs to be addressed.  Not quite sure what is happening there if you renamed the appveyor file but the system didn't pick it up.

.travis.yml
pzychotic: Maybe it's just an unfortunately display of the truncated diff here in the Conversation view or I don't understand your point.

The line changed from 'cd build_bench' to 'cd benchmark', so the directory layout would be:
build/test
build/benchmark

This layout would be automatically created by CMake with the same directory names as the source directories.

CMakeLists.txt
pzychotic: This could be put inside CMake options like EASTL_BUILD_TESTS and EASTL_BUILD_BENCHMARK and only be referenced when these options are enabled.

benchmark/CMakeLists.txt
pzychotic: Only skips the detection for a C compiler and just checks for a working C++ compiler.
Probably just my personal taste than any real benefit.

test/packages/EAStdC/CMakeLists.txt
pzychotic: Glob has the problem that any added, deleted or renamed files are not picked up when you build the project as long as you don't explicitly call CMake to regenerate your project / makefiles.
Even CMake itself discourages this practice: https://cmake.org/cmake/help/v3.1/command/file.html

Still, your call. ;)

pzychotic:
I replied to all your comments, happy to push some new commits when I know what directions you prefer.

I didn't rename the appveyor file. As far asI understand this, it's because your renamed the appveyor file to have the dot in front and didn't changed the AppVeyor settings as stated in issue #34. I suppose this should work when issue #34 gets resolved and I push a new commit.

benchmark/CMakeLists.txt
rparolin: That's cool.  We should definitely do that then.

CMakeLists.txt
rparolin: I'm on the fence about this.  I prefer if people already using EASTL don't get broken by this change and I don't see a compelling reason to change it.

.travis.yml
rparolin: Ah.  Sorry my mistaken then.  I wanted to take the opportunity to isolate all the build files under a single 'build' folder... if it already does that then great.

test/packages/EAStdC/CMakeLists.txt
rparolin: I'd prefer to leave it as is for now.

rparolin:
Updated the appveyor settings so the .appveyor.yml file is getting picked up.  I re-ran your PR and it went green now.

rparolin:
```
C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\include\xutility(2810): warning C4996: 'std::_Fill_n': Function call with parameters that may be unsafe - this call relies on the caller to check that the passed values are correct. To disable this warning, use -D_SCL_SECURE_NO_WARNINGS. See documentation on how to use Visual C++ 'Checked Iterators' [C:\projects\eastl\build\benchmark\EASTLBenchmarks.vcxproj]
  C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\include\xutility(2797): note: see declaration of 'std::_Fill_n'
  C:\projects\eastl\benchmark\source\BenchmarkAlgorithm.cpp(400): note: see reference to function template instantiation '_OutIt std::fill_n<Iterator,int,Value>(_OutIt,_Diff,const _Ty &)' being compiled
          with
          [
              _OutIt=char *,
              Iterator=char *,
              Value=char,
              _Diff=int,
              _Ty=char
          ]
  C:\projects\eastl\benchmark\source\BenchmarkAlgorithm.cpp(880): note: see reference to function template instantiation 'void `anonymous-namespace'::TestFillNStd<char*,char>(EA::StdC::Stopwatch &,Iterator,int,const Value &)' being compiled
```
https://ci.appveyor.com/project/rparolin/eastl/build/1.0.18
I noticed the tests seem to be missing the '_SCL_SECURE_NO_WARNINGS' definition causing warnings to be thrown.

CMakeLists.txt
pzychotic: So what speaks against that options and turning them off by default? No one would see the tests and benchmark pop up in their project unless explicitly enabled.
The point of this PR was to get everything build from the top CMake file and don't have to run CMake from different sub-directories to get the tests and benchmark to build. There needs to be some kind of add_subdirectory call so CMake will process the build files in these directories.

I might miss the point of your concern, so please elaborate if possible.

pzychotic:
I pushed a commit going back to glob style file listings.
I replied to the one comment I'm not sure about yet.

Concerning the SCL_SECURE thing, there never was a define for it anywhere in your project. The offending source file BenchmarkAlgorithm.cpp has a pragma to turn off the specific warning. And I can't find any reason at all why this doesn't work anymore. If that pragma is moved up right in front of #include "EASTLTest.h" it works.

CMakeLists.txt
rparolin: It's a breaking change to any using the library.  I'd be willing to do that if this is more inline with what most C++ libraries do.  I don't want to be constantly changing the build scripts back and forth.
That being said, that using options would make scripting the build / running of the benchmarks/unit tests would be easier (not require directory hopping).

pzychotic: Looks like we're getting stuck on this one. I have currently no idea how to solve this differently.

rparolin: If you add the cmake options you point out above I'll merge the PR.

>  This could be put inside CMake options like EASTL_BUILD_TESTS and EASTL_BUILD_BENCHMARK and only be referenced when these options are enabled.

Ensure you update the CI scripts and the contributing docs .  Thank you.

pzychotic: Sorry for the misunderstanding! It didn't sounded like you were convinced to me.
Everything should be in place now, hopefully.

rparolin:
Awesome.  Thank you!  :)

PR #35, opened 2016-03-14
Close varargs before reassignment with va_copy and after using them

pzychotic:
arguments wasn't closed with va_end() before being reassigned with va_copy() and argumentsSaved wasn't closed at all at the end.


rparolin:
Looks good!  Thank you.

PR #37, opened 2016-03-21
Support gcc and clang (MingW/LLVM's VS wrapper)

DragoonX6:
Now builds with GCC and Clang on windows.
Tested platforms and compilers:
* Windows
    * MSYS2 GCC 5.3.0
    * TDM-GCC 5.1.0
    * MSYS2 Clang 3.8.0
* Arch Linux
    * GCC 5.3.0
    * Clang 3.7.1

I tested MSVC before in my VM, but I haven't run my latest changes on it. I suspect it to work, though.


test/packages/EATest/CMakeLists.txt
```diff
@@ -19,10 +19,19 @@ add_definitions(-D_CRT_SECURE_NO_WARNINGS)
...
+if( UNIX AND "${CMAKE_CXX_COMPILER_ID}" STREQUAL "Intel" )
+  set( CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -fasm-blocks" )
 endif()
+
+if (CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+	SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11")
+endif()
+
+if (CMAKE_CXX_COMPILER_ID MATCHES "Clang" OR CMAKE_CXX_COMPILER_ID MATCHES "GNU")
+	set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -D_DEBUG")
```
rparolin: Enabling debug only for clang or gcc regardless of config?

test/CMakeLists.txt
```diff
@@ -19,8 +19,18 @@ include_directories(source)
...
+endif()
+
+if (CMAKE_CXX_COMPILER_ID MATCHES "Clang" OR CMAKE_CXX_COMPILER_ID MATCHES "GNU")
+	set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -D_DEBUG")
+	set(CMAKE_C_FLAGS_DEBUG "${CMAKE_C_FLAGS_DEBUG} -D_DEBUG")
+endif()
+
+if (CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+	# Disable some (maybe useless) warnings
+	set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-shift-negative-value -Wno-pointer-bool-conversion")
```
rparolin: maybe mark this as a todo.  I'd like to enable the negative shift warning... so I'm curious what code is triggering that.

include/EASTL/internal/config.h
```diff
@@ -1763,11 +1763,11 @@ typedef EASTL_SSIZE_T eastl_ssize_t; // Signed version of eastl_size_t. Concept
...
 /// This is enabled when EASTL is building built in an "open source" mode.  Which is a mode that eliminates code
 /// dependencies on other technologies that have not been released publically.
-/// EASTL_OPENSOURCE = 0, is the default.  
-/// EASTL_OPENSOURCE = 1, utilizes technologies that not publically available.
+/// EASTL_OPENSOURCE = 0, utilizes technologies that not publically available.
+/// EASTL_OPENSOURCE = 1, is the default.
 /// 
 #ifndef EASTL_OPENSOURCE
-	#define EASTL_OPENSOURCE 0
+	#define EASTL_OPENSOURCE 1
```
rparolin: EASTL_OPENSOURCE has to disable code by default if its not defined or else it will break our internal builds.

DragoonX6: Good point, this shouldn't be needed anymore now that the CMakelists are fixed.

test/packages/EATest/CMakeLists.txt
DragoonX6: CMAKE_CXX_FLAGS_DEBUG is only used when using the Debug configuration.
This define is needed to get EA_DEBUG to function correctly, as far as I know.

DragoonX6: Just checked, _DEBUG is correctly picked up by Qt Creator and EASTL_DEBUG functions as expected. If this is not picked up correctly by other IDEs, I suppose we could make the change.

rparolin: Ah, didn't realize that was limited to only debug builds.

test/CMakeLists.txt
```diff
@@ -42,10 +52,13 @@ add_definitions(-DEASTL_THREAD_SUPPORT_AVAILABLE=0)
...
 add_executable(EASTLTest ${SOURCES})
 
+set(THREADS_PREFER_PTHREAD_FLAG ON)
+find_package(Threads REQUIRED)
+
 set(EASTLTest_Libraries
     EASTL
     EATest)
-target_link_libraries(EASTLTest ${EASTLTest_Libraries})
+target_link_libraries(EASTLTest ${EASTLTest_Libraries} Threads::Threads)
```
rparolin: Can you provide more details about why this is required?

DragoonX6: std::this_thread::sleep_for in eathread.h calls nanosleep internally. nanosleep is provided by pthreads.

include/EASTL/bitset.h
```diff
@@ -1805,7 +1805,7 @@ EA_RESTORE_GCC_WARNING()
 		: base_type(value)
 	{
 		if((N & kBitsPerWordMask) || (N == 0)) // If there are any high bits to clear... (If we didn't have this check, then the code below would do the wrong thing when N == 32.
-			mWord[kWordCount - 1] &= ~(~static_cast<word_type>(0) << (N & kBitsPerWordMask)); // This clears any high unused bits.
+            mWord[kWordCount - 1] &= ~(static_cast<word_type>(~0) << (N & kBitsPerWordMask)); // This clears any high unused bits.
```
rparolin: As discussed in chat, its preferable to disable the warning here instead of introducing this change.

DragoonX6: Were we going to keep it like this? Or were we going to wrap the original cast in another cast?
As far as I know, both expressions would be identical.

rparolin: after the discovery of the integral promotion to a signed value... i think its best to wrap it with a cast.

test/packages/EATest/source/EATest.cpp
```diff
@@ -36,7 +36,7 @@
         #include <Windows.h>
         extern "C" WINBASEAPI BOOL WINAPI IsDebuggerPresent();
 
-    #if EA_WINAPI_FAMILY_PARTITION(EA_WINAPI_PARTITION_DESKTOP)
+	#if EA_WINAPI_FAMILY_PARTITION(EA_WINAPI_PARTITION_DESKTOP) && !defined(EA_COMPILER_CLANG)
```
rparolin: Fix indentation.

test/source/EASTLTestAllocator.h
```diff
@@ -9,7 +9,8 @@
 #include <EABase/eabase.h>
 #include <new>
 
-	void* operator new(size_t const size);
+	void* operator new(size_t size);
```
rparolin: Fix indentation.

test/source/TestString.inl
```diff
@@ -1163,7 +1163,7 @@ int TEST_STRING_NAME()
 		typename StringType::value_type buf[64];
 
 		StringType str(LITERAL("abcdefghijklmnopqrstuvwxyz"));
-		str.copy(buf, 10, sizeof buf);
+		str.copy(buf, 10, 0);
```
rparolin: I think this is still wrong.  
str.copy(buf, 10, 10);  // std::basic_string(ptr, size, position)

rparolin:
A few really small things but its almost ready to go in.

The eastl::string unit was interesting.  The old code was wrong, so I don't know why it wasn't failing.

test/source/TestString.inl
```diff
@@ -1163,7 +1163,7 @@ int TEST_STRING_NAME()
...
 
 		VERIFY(EA::StdC::Memcmp(buf, LITERAL("klmnopqrs"), 10) != 0);
```
rparolin: VERIFY(EA::StdC::Memcmp(buf, LITERAL("klmnopqrs"), 10) != 0);  // as written that says ensure the blocks are different. D'oh.

VERIFY(EA::StdC::Memcmp(buf, LITERAL("klmnopqrs"), 10) == 0);  // should be.

test/source/TestString.inl
DragoonX6: Pretty sure that's still wrong, as the verify statement has a 9 character literal but compares 10 characters.

rparolin: Yeah.  I fixed in another branch.  Will merge soon.

test/source/TestString.inl
```diff
@@ -1163,9 +1163,9 @@ int TEST_STRING_NAME()
 		typename StringType::value_type buf[64];
 
 		StringType str(LITERAL("abcdefghijklmnopqrstuvwxyz"));
-		str.copy(buf, 10, sizeof buf);
+        str.copy(buf, 10, 10);
```
rparolin: Please align the indentation.

test/source/TestString.inl
```diff
@@ -1163,9 +1163,9 @@ int TEST_STRING_NAME()
...
 
-		VERIFY(EA::StdC::Memcmp(buf, LITERAL("klmnopqrs"), 10) != 0);
+        VERIFY(EA::StdC::Memcmp(buf, LITERAL("klmnopqrst"), 10) == 0);
```
rparolin: Please align the indentation.

test/source/GetTypeName.h
```diff
@@ -81,7 +81,12 @@ eastl::string GetTypeName()
 			const char* pDemangledName = NULL;
 		#endif
 
-		if(pDemangledName && pDemangledName[0])
+		// Gets rid of pointer to bool conversion warning when pDemangledName is an array.
+		#if EASTL_LIBSTDCPP_DEMANGLE_AVAILABLE
+			if(pDemangledName && pDemangledName[0])
```
rparolin: I'd prefer to not include this delta as the added code complexity isn't required to optimize away an additional comparison.

test/source/GetTypeName.h
```diff
@@ -97,12 +102,12 @@ eastl::string GetTypeName()
 		else if(eastl::is_rvalue_reference<T>::value)
 			result += "&&";
 
+		#if EASTL_LIBSTDCPP_DEMANGLE_AVAILABLE
```
rparolin: Same here.   I'd prefer to leave it as is.

include/EASTL/type_traits.h
```diff
@@ -865,7 +865,10 @@ namespace eastl
 	// 
 	///////////////////////////////////////////////////////////////////////
 
-	#if defined(EA_COMPILER_NO_VARIADIC_TEMPLATES) || defined(EA_COMPILER_MSVC) // VS2013 fails to compile the variadic code below due to what looks like a deficiency in their handling of integral variadic template parameters.
+	// VS2013 fails to compile the variadic code below due to what looks like a deficiency in their handling of integral variadic template parameters.
+	// Also mingw's clang and gcc fail to compile the code (on windows, linux untested).
+	#if defined(EA_COMPILER_NO_VARIADIC_TEMPLATES) || defined(EA_COMPILER_MSVC) || \
```
rparolin: Since I flagged other edits in this PR...If you could place this all on a single line that would be preferable.

Thanks.

rparolin:
Accepted!  Thank you for all the work.

PR #38, opened 2016-03-22
Allow libs of different configs to be installed in parallel

### LINKED ISSUE #36: CMake install overwrites libs of different configs
@Leandros introduced a CMake install command in 8be33d4ba8ecefadd3bfde95f9f980523f64882b that can't install the EASTL.lib for different configs like Debug and Release in parallel. The last install will always overwrite any other lib because of the same name.

There are a few ways to make this possible in CMake.

One is to use `set(CMAKE_DEBUG_POSTFIX "d")` or `set_property(TARGET EASTL PROPERTY DEBUG_POSTFIX "d")` to change the name of the lib in Debug config to EASTLd.lib. Not a big fan of this and a breaking change for anyone using the Debug version. Beware `set(CMAKE_DEBUG_POSTFIX "d")`would change the postfix for **all** libs.

The other option would be to install the libs into sub-directories like:
lib/Debug
lib/Release

The modern CMake approach would be to use generator expressions like:
`install(TARGETS EASTL DESTINATION lib/$<LOWER_CASE:$<CONFIG>>)`
This would also allow transformations of the directory name to lower case or anything else generator expressions can do. Unfortunately the install(TARGET...) command only supports generator expressions with CMake v3.3+.

If increasing the minimum required CMake version is not an option, passing the BUILD_TYPE escaped to the install script (note the backslash in front of the ${...}) will also work:
`install(TARGETS EASTL DESTINATION lib/\${BUILD_TYPE})`

A trivial change to make yourself in any case, if there are no objections. Or let me know if you prefer a PR for one of the versions.

pzychotic:
Fixes #36.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -69,7 +69,7 @@ target_link_libraries(EASTL EABase)
 #-------------------------------------------------------------------------------------------
 # Installation
 #-------------------------------------------------------------------------------------------
-install(TARGETS EASTL DESTINATION lib)
+install(TARGETS EASTL DESTINATION lib/\${BUILD_TYPE})
 install(DIRECTORY include/EASTL DESTINATION include)
 install(DIRECTORY test/packages/EABase/include/Common/EABase DESTINATION include)
 


rparolin:
Thank you.

PR #41, opened 2016-04-07
Fixed issue where cmake (3.2.2) does not understand its own tests 

HFTrader:
CMake Error at /usr/local/share/cmake-3.2/Modules/CheckIncludeFiles.cmake:74 (try_compile):
  Unknown extension ".c" for file

Adding the C language to the "project" command causes other problems.
The solution was to enable the "C" language explicitly in CMakeLists.txt
    enable_language(C)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,6 +3,7 @@
 #-------------------------------------------------------------------------------------------
 cmake_minimum_required(VERSION 3.1)
 project(EASTL CXX)
+enable_language(C)
 
 #-------------------------------------------------------------------------------------------
 # Options
@@ -83,4 +84,3 @@ target_link_libraries(EASTL EABase)
 install(TARGETS EASTL DESTINATION lib)
 install(DIRECTORY include/EASTL DESTINATION include)
 install(DIRECTORY test/packages/EABase/include/Common/EABase DESTINATION include)
-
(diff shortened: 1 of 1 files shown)


rparolin:
How are you generating this error?

HFTrader:
Well I am running Ubuntu 15.04 and because cmake is so old on that distro, I installed it from a newer version (3.2.2) then is when I had that problem. But reinstalling the newer 3.5 the problem goes away.

rparolin:
In a previous PR, the following was introduced to tell cmake it was always building C++ code.  I don't know where the *.c files would be coming from?  Interesting that it goes away with cmake-3.5.

```
cmake_minimum_required(VERSION 3.1)
  project(EASTL CXX)       // calls 'enable_language(CXX)' internally
 +enable_language(C) 
```

HFTrader:
Well looking at the logs, it comes when trying to compile internally some .c files to check for flags etc. 
I think this might be a known issue when you depart from the default (C/C++). 
https://cmake.org/pipermail/cmake/2007-May/013976.html
When you ask for C++ only, as in this project, it disables the C language and then shoots itself in the foot. Perhaps it should also know it should create test files with the cpp extension instead. Go figure.
It seems to be fixed in the latest cmake but there are plenty of cmake 2.8-3.2 lingering around. It's up to you to add this protection.

rparolin:
Since the default is C and C++, maybe we should just drop the 'CXX' and revert it back to:
```
cmake_minimum_required(VERSION 3.1)
  project(EASTL)       // by default C and CXX
```

HFTrader:
I think that actually makes much more sense.

rparolin:
Cool.  Feel free to update this PR and I'll accept it.  Thanks!

HFTrader:
I checked in the update as discussed. Thanks!

rparolin:
Merged.  Thank you!

PR #49, opened 2016-08-14
Fix linking error on windows when not using msvc.

DragoonX6:
No compiler that I know of on Windows has librt, as such we shouldn't include it.
Verified working.

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -52,7 +52,7 @@ set(THREADS_PREFER_PTHREAD_FLAG ON)
 find_package(Threads REQUIRED)
 
 set(EASTLTest_Libraries EASTL EATest)
-if((NOT APPLE) AND (NOT MSVC))
+if((NOT APPLE) AND (NOT WIN32))
 	target_link_libraries(EASTLTest ${EASTLTest_Libraries} Threads::Threads rt)
 else()
 	target_link_libraries(EASTLTest ${EASTLTest_Libraries} Threads::Threads)


DragoonX6:
Alternatively, instead of checking if it's not Apple or Windows, we could just check if it's a Unix or BSD variant. I don't know if consoles have this though.

rparolin:
We can't test consoles publicly and the cmake scripts are not used
internally.  So I don't think it's major concern.

I was trying to accommodate gcc/clang running on Windows by limiting the
expression to MSVC.

DragoonX6:
Alright, so I guess this is ready to be merged?

rparolin:
Sorry, I don't see how changing it to Win32 is preferred?  MSVC covers all cases in my mind.  Win64?  Clang on Windows? GCC on Windows?  Basically we don't want to use pthreads if the compiler is MSVC.  How is this problematic?

DragoonX6:
The line `target_link_libraries(EASTLTest ${EASTLTest_Libraries} Threads::Threads)` tells cmake to link to threads, with preference to pthreads if it's available, which with MSVC is not, so this isn't a problematic line for MSVC.

However, the line 'target_link_libraries(EASTLTest ${EASTLTest_Libraries} Threads::Threads rt)' tells it to link to both threads and librt, which is a Unix only library, so Cygwin, Linux, and FreeBSD as far as I'm aware.

Thus checking only for MSVC, instead of Windows entirely (WIN32 means both 32 and 64 bit Windows builds), will make the build fail since the linker can't find librt.

rparolin:
Ah, my mistake.  I'll merge that in.  Thanks!

DragoonX6:
No problem, thank you too!

PR #51, opened 2016-08-19
Add MinGW-w64 (MSYS2) CI scripts

DragoonX6:
This PR adds scripts for appveyor to automatically check mingw-w64 (MSYS2) compatibility.
Some things I did:
- Switch GCC to C++14, since it needs to run in C++14 mode for variadic template support.
- Add C++14 deleters to the test allocators.
- Turn off strict aliasing on GCC, without this the intrusive list test won't pass. (Probably needs some investigating to make sure this also works on compilers that do not have such an option.)
- Clean up the CMakeLists a bit.

The mingw-w64 appveyor build script is separate from the standard one, I've done this in an attempt to shorten the amount of time before you can see if integration failed.
Another thing I did, I stopped clang from building in release mode in x86 configurations, since it ICEs on about 90% of the source code.


test/source/EASTLTestAllocator.cpp
```diff
@@ -372,25 +372,32 @@
...
-	void* operator new[](size_t size, size_t alignment) 
+	void* operator new[](size_t size, size_t alignment)
 		{ return Internal::EASTLAlignedAlloc(size, alignment); }
 
 	void* operator new[](size_t size, size_t alignment, const std::nothrow_t&)EA_THROW_SPEC_NEW_NONE()
 		{ return Internal::EASTLAlignedAlloc(size, alignment); }
 
+	// C++14 deleter
+	void operator delete(void* p, std::size_t sz ) EA_THROW_SPEC_DELETE_NONE()
+	{ Internal::EASTLAlignedFree(p); (void)sz; }
```
rparolin: prefer EA_UNUSED(sz); here instead of '(void)sz;'

test/packages/EABase/include/Common/EABase/eahave.h
```diff
@@ -559,7 +559,7 @@
 #endif
 
 #if !defined(EA_HAVE_ISNAN) && !defined(EA_NO_HAVE_ISNAN)
-	#if defined(EA_PLATFORM_MICROSOFT) && !defined(EA_COMPILER_CLANG)
+	#if defined(EA_PLATFORM_MICROSOFT) && !defined(EA_PLATFORM_MINGW)
```
rparolin: We are removing clang from the expression?

rparolin:
- The travis CI builds are failing on this PR to due the '-std=c++14' flag that needs to be resolved before it can be accepted.
- EASTL must provide backwards compatibility for a wide variety of code.  Deprecation of features or language support (ie. no longer supporting C++11) must be agreed on.  In addition, a grace period of backwards compatibility is required.  We can definitely enable a separate CI build step that builds GCC/Clang in C++14 mode but the C++11 build can't be disabled and we have to continue supporting that line as a lot of users (including internally) are building on C++11 compilers.  
-In the case of variadic templates, we generally use the preprocessor to remove the pieces that depend on variadics or provide a manual unrolling up to some number N of parameters, if necessary.  This is less important as support for a compiler that does not support variadics has recently been sunset internally but I thought I'd mention it.
- General feedback for pull requests.  Avoid whitespace changes with other larger code changes or infrastructure changes as it distracts the reviewer by obscuring the important pieces of the review.
- ci-build.cmd / ci-test.cmd I'd prefer if all that code was just in a travis yml file.  Especially since the other yml files don't defer to other helper files.  
- Can you explain why we need to add a dependency on ninja?

Thanks you for the PR.  I look forward to seeing the updates.

test/packages/EABase/include/Common/EABase/eahave.h
DragoonX6: GCC also failed on that line, both Mingw's clang and gcc expose EA_PLATFORM_MINGW.

DragoonX6:
> General feedback for pull requests. Avoid whitespace changes with other larger code changes or infrastructure changes as it distracts the reviewer by obscuring the important pieces of the review.

Yeah, sorry, Qt Creator automatically changes inconsistent whitespace on save, I'll look in turning that off.

>The travis CI builds are failing on this PR to due the '-std=c++14' flag that needs to be resolved before it can be accepted.

Alright, I'll add checks for GCC and Clang versions.

>ci-build.cmd / ci-test.cmd I'd prefer if all that code was just in a travis yml file. Especially since the other yml files don't defer to other helper files.

I had to do this because of a limitation in appveyor, where it interprets multiple command lines as a single one, causing it to error out during build.

>Can you explain why we need to add a dependency on ninja?

Not a dependency, I just added it as a preference, since it builds faster and automatically picks the right amount of cores during compilation. Changing -GNinja to -GMSYS Makefiles or any other generator that uses Makefiles just works as you would expect.

test/source/EASTLTestAllocator.cpp
DragoonX6: Done.

DragoonX6:
I fixed the problems with the PR, minimized the white space changes and squashed them together for the sake of readability.

DragoonX6:
Squashed and pushed, I hope it works now.

benchmark/CMakeLists.txt
```diff
@@ -15,44 +15,83 @@ include(CTest)
...
 include_directories(../test/source)
 
 #-------------------------------------------------------------------------------------------
 # Compiler Flags
 #-------------------------------------------------------------------------------------------
-set(CMAKE_CXX_STANDARD 11)
-
-if (CMAKE_CXX_COMPILER_ID MATCHES "Clang")
-	SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11")
+if(CMAKE_CXX_COMPILER_ID MATCHES "Clang" AND NOT APPLE) # Apple clang uses different versioning.
```
rparolin: Is it possible to have this code in a single file and reference it in the various cmakelist.txt files?  That way we don't have to repeat ourselves and make future changes in multiple places.

DragoonX6: I'm pretty sure it should be possible, I'll look into that this weekend.

test/CMakeLists.txt
```diff
@@ -3,24 +3,72 @@
...
+# Include directories
 #-------------------------------------------------------------------------------------------
-include_directories(source)
+include_directories(source "../include")
 
 #-------------------------------------------------------------------------------------------
 # Compiler Flags
 #-------------------------------------------------------------------------------------------
-set(CMAKE_CXX_STANDARD 11)
+if(CMAKE_CXX_COMPILER_ID MATCHES "Clang" AND NOT APPLE) # Apple clang uses different versioning.
```
rparolin: Can we not repeat ourselves here?  Same comment as above.

test/packages/EABase/include/Common/EABase/eahave.h
rparolin: MinGW isn't the only way clang can run on Windows.  We'd need MinGW have to be a seperate case in this block... similar to how Cygwin is handled.

test/packages/EAStdC/CMakeLists.txt
```diff
@@ -21,25 +21,63 @@ add_definitions(-D_CHAR16T)
 # Compiler Flags
 #-------------------------------------------------------------------------------------------
 if( UNIX AND "${CMAKE_CXX_COMPILER_ID}" STREQUAL "Intel" )
-  set( CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -fasm-blocks" )
+    set( CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -fasm-blocks" )
 endif()
 
-if (CMAKE_CXX_COMPILER_ID MATCHES "Clang")
-	SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11")
+if(CMAKE_CXX_COMPILER_ID MATCHES "Clang" AND NOT APPLE) # Apple clang uses different versioning.
```
rparolin: Same comment as above.  Lets find a common place for this.

test/packages/EATest/CMakeLists.txt
```diff
@@ -20,16 +20,54 @@ add_definitions(-D_CRT_SECURE_NO_WARNINGS)
 # Compiler Flags
 #-------------------------------------------------------------------------------------------
 if( UNIX AND "${CMAKE_CXX_COMPILER_ID}" STREQUAL "Intel" )
-  set( CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -fasm-blocks" )
+    set( CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -fasm-blocks" )
 endif()
 
-if (CMAKE_CXX_COMPILER_ID MATCHES "Clang")
-	SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11")
+if(CMAKE_CXX_COMPILER_ID MATCHES "Clang" AND NOT APPLE) # Apple clang uses different versioning.
```
rparolin: Same comment as above.

.appveyor-mingw.yml
```diff
@@ -0,0 +1,18 @@
+version: 1.0.{build}
```
rparolin: how does appveyor know to run this yml file?  I looked at the CI builds attached to this PR and it doesn't appear to be running it.  is there an appveyor run that I can see executing the builds described by this file?

rparolin:
New comments added.  Thanks!

.appveyor-mingw.yml
DragoonX6: You would have to add an extra appveyor build for this. I have done that in an attempt to shorten the overall build time. But I have an idea on how I can put it in one file and still keep the overall build time low.

rparolin: I don't have an issue with the MingW builds being a seperate yml file.  I guess I figured the PR would be running these builds.  I guess I'd have to set it up first it won't just detect that it needs those builds like travis would.

test/packages/EABase/include/Common/EABase/eahave.h
DragoonX6: https://github.com/electronicarts/EASTL/blob/master/test/packages/EABase/include/Common/EABase/eahave.h#L562

Since both Mingw's clang and Mingw's gcc define EA_PLATFORM_MICROSOFT, but don't have the _isnan and _finite symbols, we should remove mingw in its entirety from the expression.
The reason I added EA_COMPILER_CLANG earlier on that line is because I didn't test for MingW's gcc at the time.

Since MingW exposes isnan and isinf, we can remove win clang from the apple expression, since std::isnan is just `namespace std { using ::isnan; }` anyway.

I'm pretty sure that would work for clang-cl since that uses the msvc libraries.

Thus I still think the expression how it is now is correct, but that we should roll back the apple expression to being purely apple again.

DragoonX6: I updated the expression to fallback to the last case when using mingw.

DragoonX6:
Updated the PR with the requested changes, hope it's good to now.

rparolin:
I think this looks good.  Where can I see the MinGW builds running on AppVeyor?

DragoonX6:
You can view the build logs here: https://ci.appveyor.com/project/DragoonX6/eastl-oi7o0

rparolin:
Only the MingW builds are shown on this page.  When I merge the PR do I
have point the EASTL appveyor to your new yml file?  Will it get picked
automatically?  Does the site support specifying two files?

DragoonX6:
What I did was add a new project on appveyor called EASTL-MinGW and point it to the corresponding yml file. Again, if you want it picked up automatically you have to change the behavior of the current yml file.

rparolin:
What specifically would have to change?  I'm leaning towards just having a
single file.  I'm not super concerned about the overall build time.  What
times were you seeing?  In addition, I noticed there is a lot of files
migrating to the root.  Can you move the new  cmake and cmd files you added
to the scripts folder?

DragoonX6:
Some behavior would need to be added to the current yml file to split mingw builds from VS builds.
The time is increased about by 4 minutes per build if you don't split the builds via either some extra behavior or splitting the files.
What scripts folder are you talking about? From my knowledge it's standard CMake practice to add a folder called CMake to the root of the project which contains shared/split behavior.
I guess we could move the cmd files to a folder called appveyor?

rparolin:
having a folder at the root called "scripts" is an internal standard.  any scripts files could be migrated there.

DragoonX6:
Okay, I updated the PR to comply with your requests.
I also found a bug in the appveyor script, which I subsequently fixed.
The bug in question was that 64 bit msvc builds were never made.

Anyway, I hope it's alright to merge now.

DragoonX6:
I noticed something else, we only test msvc release builds, is this intended?

rparolin:
Looks good.  Thank you!

PR #52, opened 2016-08-24
Changed insertion_sort so wont decrement an iterator too far.

Jonathan-Weinstein:
Modified the two versions of insertion_sort so that they wont decrement a pointer to a potentially undefined/invalid state.

Also, though it might not get used much,  I left a new version of merge_sort/stable_sort in comments as a proposal.
It only needs the buffer to be at least half the input size, instead of equal to it, and it seems to be a little faster from some testing I did.

### DIFF
--- a/include/EASTL/InsertionSortDriver.cpp
+++ b/include/EASTL/InsertionSortDriver.cpp
@@ -1,0 +1,178 @@
+#include <stdio.h>
+#include <stdint.h>
+#include <time.h>
+#include <assert.h>
+typedef unsigned long ulong;
+typedef int32_t i32;
+
+#ifdef EASTL_VALIDATE_COMPARE
+    #undef EASTL_VALIDATE_COMPARE
+    #define EASTL_VALIDATE_COMPARE(x) ((void)0)
+#endif // EASTL_VALIDATE_COMPARE
+
+#include <EASTL/sort.h>
+
+const auto Less=[](i32 a, i32 b){ return a<b; };
+
+enum:size_t{OneMil=10000000u};
+
+#ifdef _WIN32
+#define WIN32_LEAN_AND_MEAN
+#include <windows.h>
+#include <wincrypt.h>
+bool FillRandom(void* a, size_t bytes)
+{
+    HCRYPTPROV   hCryptProv;
+
+    if (CryptAcquireContext(&hCryptProv, nullptr, nullptr, PROV_RSA_FULL, CRYPT_VERIFYCONTEXT))
+    {
+        const bool ret=CryptGenRandom(hCryptProv, bytes, static_cast<BYTE*>(a));
+
+        puts(ret ? "Random sequence successfully generated" : "Error during CryptGenRandom");
+
+        return CryptReleaseContext(hCryptProv, 0) ? (puts("Crypt handle released"), ret) : (puts("Crypt handle failed to release"), false);
+    }
+    else
+    {
+        puts("A cryptographic service handle could not be acquired");
+        return false;
+    }
+}
+#else
+#include <fcntl.h>
+#include <unistd.h>
+bool FillRandom(void* a, size_t bytes)
+{
+    const int fd=open("/dev/random", O_RDONLY);
+    if (fd>=0)
+    {
+        const ptrdiff_t nRead=read(fd, a, bytes);
+        bool ret = (nRead<0 || size_t(nRead)<bytes) ? (perror("read"), false) : true;
+        if (close(fd)!=0) perror("close"), ret=false;
+        return ret;
+    }
+    else
+    {
+        perror("open() failed on /dev/random");
+        return false;
+    }
+}
+#endif//WIN32
+
+void TestNewInsertionSort(const int NumTests=30)
+{
+    enum:size_t{SmallArraySize=20u, N=OneMil};
+
+    static i32 data[N];
+
+    FILE* TempDataFile=tmpfile();
+    if (TempDataFile==NULL)
+    {
+        perror("couldn't create temp file");
+        return;
+    }
+
+    if (!FillRandom(data, sizeof data)) { return; }
+    if (fwrite(data, 4, N, TempDataFile)!=N) { perror("fwrite"); return; }
+
+    int observe=0;
+    clock_t start, stop, oldSum=0, newSum=0;
+    i32* it=NULL;
+    for (int loops=NumTests; loops-- > 0; )
+    {
+        //testing current version
+        rewind(TempDataFile);
+        if (fread(data, 4, N, TempDataFile)!=N) { perror("fread"); return; }
+
+        start=clock();
+        //test many small arrays instead of a big one, or will take a while
+        it=data;
+        for (size_t j=SmallArraySize; j<=N; it+=SmallArraySize, j+=SmallArraySize)
+        {
+            eastl::insertion_sort(it, data+j);
+            //assert(eastl::is_sorted(it, data+j));
+            observe ^= *it;
+        }
+        stop=clock();
+        oldSum += stop-start;
+
+        //testing new version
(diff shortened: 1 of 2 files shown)


rparolin:
Thanks for providing a PR for this todo.  I''ll review the change and provide feedback.  Stay tuned.

include/EASTL/sort.h
```diff
@@ -205,77 +205,144 @@ namespace eastl
...
+	 *
+	 * An improvement for larger inputs would be to use upper_bound to find the insertion position,
+	 * followed by memmove, but it will still be slow for large inputs, and not worth it for small ones.
+	 *
+	 * If base[-1] is a sentinel the short-circuit can be removed, but this is a general purpose version.
+	 */
+	
+	//compare functor version
+	template<class Iter, class Compare>//Bidirectional Iterator, though something like a list shouldn't use this
+	void insertion_sort(Iter const base, Iter const past, Compare cmp)
```
rparolin: The template parameter names and function parameters are named fairly consistently across the code base.  Please avoid renaming them unless it can't be avoided.

include/EASTL/sort.h
```diff
@@ -205,77 +205,144 @@ namespace eastl
...
-	/// insertion_sort
-	///
-	/// Since insertion_sort requires that the data be addressed with a BidirectionalIterator and 
-	/// not the more flexible RandomAccessIterator, we implement the sort by doing a for loop within
-	/// a for loop. If we were to specialize this for a RandomAccessIterator, we could replace the
-	/// inner for loop with a call to upper_bound, which would be faster.
-	///
-	template <typename BidirectionalIterator, typename StrictWeakOrdering>
-	void insertion_sort(BidirectionalIterator first, BidirectionalIterator last, StrictWeakOrdering compare)
+	/* Insertion Sort
```
rparolin: Please use the previous comment format.

include/EASTL/sort.h
```diff
@@ -205,77 +205,144 @@ namespace eastl
...
-				*iNext = temp;
-			}
-		}
-	} // insertion_sort
-
+	    typedef typename eastl::iterator_traits<Iter>::value_type value_type;
+	
+	    if (base!=past)
+	    {   //this could perhaps be written more idiomatic: for (T *left=base, *sorted=base+1; sorted!=past; left=sorted++)
+	        for (Iter left=base, sorted=base; ++sorted!=past; left=sorted)
```
rparolin: The format and variable names aren't in the EASTL standard.  Look at or use the .clang-format file in the repo for what to do there.

include/EASTL/sort.h
```diff
@@ -205,77 +205,144 @@ namespace eastl
...
+	        {
+	            if (*sorted < *left)
+	            {
+	                const value_type val=*sorted;//should we care about doing moves instead?
```
rparolin: Did this change have any impact on our benchmarks that you observed?

include/EASTL/sort.h
```diff
@@ -205,77 +205,144 @@ namespace eastl
...
-	void insertion_sort(BidirectionalIterator first, BidirectionalIterator last)
+	    typedef typename eastl::iterator_traits<Iter>::value_type value_type;
+	
+	    if (base!=past)
+	    {   //this could perhaps be written more idiomatic: for (T *left=base, *sorted=base+1; sorted!=past; left=sorted++)
+	        for (Iter left=base, sorted=base; ++sorted!=past; left=sorted)
+	        {
+	            if (cmp(*sorted, *left))
+	            {
+	                const value_type val=*sorted;//should we care about doing moves instead?
```
rparolin: Was there a reason why you changed from 'const value_type temp(*iSorted);' an copy-assignment expression?  I believe in practice these should be equivalent as (I believe) the standard this will always be directed-initialization (ie. not a copy-constructed or copy-assigned).  I was curious if there was a reason for the change?  If there wasn't a problem you were working around I'd prefer to keep the behaviour as is to avoid any subtle knock-on affects.

include/EASTL/sort.h
```diff
@@ -205,77 +205,144 @@ namespace eastl
...
-
-	template <typename BidirectionalIterator>
-	void insertion_sort(BidirectionalIterator first, BidirectionalIterator last)
+	    typedef typename eastl::iterator_traits<Iter>::value_type value_type;
+	
+	    if (base!=past)
+	    {   //this could perhaps be written more idiomatic: for (T *left=base, *sorted=base+1; sorted!=past; left=sorted++)
+	        for (Iter left=base, sorted=base; ++sorted!=past; left=sorted)
+	        {
+	            if (cmp(*sorted, *left))
```
rparolin: Did you observe any regression in the benchmarks?

include/EASTL/sort.h
```diff
@@ -205,77 +205,144 @@ namespace eastl
...
-	} // insertion_sort
-
-
-
-	template <typename BidirectionalIterator>
-	void insertion_sort(BidirectionalIterator first, BidirectionalIterator last)
+	    typedef typename eastl::iterator_traits<Iter>::value_type value_type;
+	
+	    if (base!=past)
+	    {   //this could perhaps be written more idiomatic: for (T *left=base, *sorted=base+1; sorted!=past; left=sorted++)
```
rparolin: Currently, I"m leaning towards the more (as you say) the more idiomatic version for code clarity.

include/EASTL/sort.h
```diff
@@ -205,77 +205,144 @@ namespace eastl
...
+	        for (Iter left=base, sorted=base; ++sorted!=past; left=sorted)
+	        {
+	            if (cmp(*sorted, *left))
+	            {
+	                const value_type val=*sorted;//should we care about doing moves instead?
+	                Iter right=sorted;
+	
+	                do
+	                {
+	                    EASTL_VALIDATE_COMPARE(!cmp(*left, val));//validate cmp is sane by checking other way
```
rparolin: As far as I can tell the algorithm looks good.  Just needs to be cleaned up from a coding standards perspective.

Jonathan-Weinstein:
Hey, thanks for giving me feedback.

I think I fixed the style changes in this new commit.
I also made a driver file to test and benchmark the new version.

Sorting many arrays of length 20 of 32-bit ints, the new one seems to run in about 0.94 the time
of the current insertion sort one on my system.

Testing in a similar fashion, I find the new merge sort I'm proposing to run in 0.82 the time of the current one, and it uses half the space. This was from sorting arrays of 1-million 32-bit ints.

My benchmarking isn't the best though, any feedback will be much appreciated.

include/EASTL/sort.h
```diff
@@ -202,80 +202,189 @@ namespace eastl
...
+	            *iRight=temp;
+	        }//outer loop
+	    }
+	}//insertion_sort (less than)
+	
+	///This following Merge Sort or something similar is worth considering compared to the current version.
+	///First, the buffer only needs a length of half the data, not the full length as it does now.
+	///Second, it starts by using Insertion Sort on small arrays, which noticeably improves the performance. The full sort is still stable.
+	
+	//the following is like a normal merge, but nothing remains to be copied if the left range is exhausted first
```
rparolin: I'd suggest taking the merge sort changes and submitting them in another PR so they can discussed separately from the insertation sort changes.

include/EASTL/InsertionSortDriver.cpp
```diff
@@ -0,0 +1,208 @@
+#include <stdio.h>
+#include <stddef.h>
+#include <stdint.h>
+#include <time.h>
```
rparolin: I think we should migrate something like to this the EASTL benchmarks instead of a completely separate driver.

Jonathan-Weinstein:
Thanks for the feedback, Ill work on both of them.

rparolin:
Thank you!

Issue #50, opened 2016-08-14
Figure out a fix for timing numbers in MingW


DragoonX6:
In the attached file [EASTLBenchmark output.txt](https://github.com/electronicarts/EASTL/files/417179/EASTLBenchmark.output.txt) you can see a lot of zeros and a lot of nans.
This is of course unacceptable when it comes to measuring any kind of performance.

An earlier [bug report on the GNU tracker](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=63400) shows that this might be related to the chrono implementation with regards to pthreads. The OP points out that using steady_clock instead of high_resolution_clock yields more precise results and I can indeed confirm this, seen in the attached file: [EASTLBenchmark steady_clock output.txt](https://github.com/electronicarts/EASTL/files/417191/EASTLBenchmark.steady_clock.output.txt)
One other interesting thing is how this is 0.06 seconds faster too.

The OP also [links to the MingW mailing list](https://sourceforge.net/p/mingw-w64/mailman/message/31672495/) where somebody notes how this issue doesn't occur when using boost's chrono implementation.

So there are a few ways on how we can fix this:
- We simply switch to steady_clock on MingW builds.

   However this feels like a band-aid fix where I'm not sure the times recorded are as accurate as MS's implementation or on the other platforms.
- We use boost's chrono instead.

   This is an undesirable solution, as this introduces a huge dependency.
- We use boost's chrono, but rip it out of boost instead of introducing it as a dependency.

   This is less undesirable depending on the size of boost's chrono and could introduce licensing issues.
- We fall back to C API entirely.

   If this can be done with standard C code, it's perhaps the most desirable of the solutions, as it is guaranteed to work on all platforms. 
- We submit a patch to libstdc++.

   This is the most ideal solution in the long run, but this means the problem still remains, unless we combine it with one of the previous mentioned solutions.

Maybe an EASTL chrono is needed?
I think if we dig some deeper, the steady_clock switch might prove sufficient, if not, I think falling back to the C API might be the most desirable solution.

rparolin:
I added chrono support to EASTL.  
https://github.com/electronicarts/EASTL/blob/master/include/EASTL/chrono.h

I believe you are refering to the EAStdC Stopwatch implementation that used the standard chrono.  EAStdC is a stub implementation that isn't used internally (fingers crossed that I can open source that soon)... We could definitely change that to use steady_clock instead of high_resolution_clock.

[...]

PR #53, opened 2016-09-06
Fix timing numbers for MinGW (issue #50).

DragoonX6:
This PR fixes issue #50.


rparolin:
What's the issue for using the real_time_clock on MinGW?

DragoonX6:
high_resolution_clock uses a less optimal clock, check out the referenced issue again for more information on that, I documented it all there.

PR #54, opened 2016-09-13
Hack to avoid warning on MSVC

ofwinterpassed:
MSVC (at least 2013) gives a warning if you store a type with an auto generated destructor in some containers (e.g. vector_set) claiming p is unreferenced. This is clearly a problem from their side but this small change ensures that the code compiles.

### DIFF
--- a/include/EASTL/memory.h
+++ b/include/EASTL/memory.h
@@ -1227,6 +1227,7 @@ namespace eastl
 	template <typename T>
 	inline void destruct(T* p)
 	{
+		(void)p;
 		p->~T();
 	}
 
--- a/include/EASTL/memory.h
+++ b/include/EASTL/memory.h
@@ -1227,7 +1227,10 @@ namespace eastl
 	template <typename T>
 	inline void destruct(T* p)
 	{
-		(void)p;
+		// https://msdn.microsoft.com/query/dev14.query?appId=Dev14IDEF1&l=EN-US&k=k(C4100)&rd=true
+		// "C4100 can also be issued when code calls a destructor on a otherwise unreferenced parameter
+		//  of primitive type. This is a limitation of the Visual C++ compiler."
+		EA_UNUSED(p);
 		p->~T();
 	}
 


rparolin:
Thanks for submitting a PR.  I'm not sure I understand the problem.  How can 'p' in this case be unreferenced when the code is explicitly calling through the pointer the destructor?  I'm not sure how the destructor being auto-generated or not is related to the problem.

ofwinterpassed:
I'll try to explain it a bit clearer: This is a hack to avoid a bug in MSVC. The bug is that if you use a templated function like the destruct function in eastl which only calls the destructor on the templated argument and the template type is a class which doesn't define the destructor MSVC (2013) will wrongly give you a warning about the argument being unreferenced. I.e. The code in eastl is perfectly valid C++ according to the standard and my change just avoids a bug in MSVC. Hope this makes it clearer.

DragoonX6:
Wouldn't it be better to ignore the MSVC warning here? What's the actual warning code, is it actually a useful warning?
I'd personally prefer we write a bit more code to disable the warning with an explanation, than a random EA_UNUSED to avoid a compiler bug, which feels less of a hack. Ultimately the decision lies with @rparolin.

rparolin:
I agree with @DragoonX6.  Without documenting why we are disabling it, it makes it fragile under future maintenance.  I'd actually like to see more information on the specific warning and the nature of the MSVC bug that is causing it.  We make heavy use of MSVC  internally and we haven't come across this issue - which surprises me.

ofwinterpassed:
I'll try to provide more specifics today. Thanks for your patience.

asRIA:
Hi, Error looks like: 
```
Severity	Code	Description	Project	File	Line	Suppression State
Warning	C4100	'p': unreferenced formal parameter (compiling source file D:\test\Foo.cpp)	test D:\test\EASTL\include\EASTL\memory.h	1228	
```

asRIA:
We found this issue when we started to use `emplace` on `vector_set`.
The simplest test case is: 
```cpp
#include <EASTL/memory.h>
struct A {
	~A() {};
};

struct B {
	~B() = default;
};

struct C {
};

class Foo {
public:
	Foo() {
		eastl::destruct<A>(0); // clear
		eastl::destruct<B>(0); // warns
		eastl::destruct<C>(0); // warns
	}
};
```

rparolin:
https://msdn.microsoft.com/query/dev14.query?appId=Dev14IDEF1&l=EN-US&k=k(C4100)&rd=true
"C4100 can also be issued when code calls a destructor on a otherwise unreferenced parameter of primitive type. This is a limitation of the Visual C++ compiler."

That's unfortunate.

include/EASTL/memory.h
```diff
@@ -1227,6 +1227,7 @@ namespace eastl
 	template <typename T>
 	inline void destruct(T* p)
 	{
+		(void)p;
```
rparolin: Please use our portable macro "EA_UNUSED" instead of casting to void.  Please reference the MSDN article in a comment saying this cast is necessary due to a limitation in MSVC.

rparolin:
Thank you for the PR.

PR #60, opened 2016-10-01
Fix aliasing issues with intrusive_list.

DragoonX6:
This _should_ fix the aliasing issues with intrusive_list on GCC that are seen in release mode. Now all node_types are stored in intrusive_list_nodes, allowing the compiler to pickup on the EASTL_MAY_ALIAS attribute/directive.

### DIFF
--- a/include/EASTL/intrusive_list.h
+++ b/include/EASTL/intrusive_list.h
@@ -791,8 +791,8 @@ namespace eastl
 				EASTL_FAIL_MSG("intrusive_list::insert(): element already on a list.");
 		#endif
 
-		node_type& next = *const_cast<node_type*>(pos.mpNode);
-		node_type& prev = *static_cast<node_type*>(next.mpPrev);
+		intrusive_list_node& next = *const_cast<node_type*>(pos.mpNode);
+		intrusive_list_node& prev = *static_cast<node_type*>(next.mpPrev);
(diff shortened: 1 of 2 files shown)


include/EASTL/intrusive_list.h
```diff
@@ -791,8 +791,8 @@ namespace eastl
 				EASTL_FAIL_MSG("intrusive_list::insert(): element already on a list.");
 		#endif
 
-		node_type& next = *const_cast<node_type*>(pos.mpNode);
-		node_type& prev = *static_cast<node_type*>(next.mpPrev);
+		intrusive_list_node& next = *const_cast<node_type*>(pos.mpNode);
```
DragoonX6: I'm not sure what you mean, do you mean line 795? I only changed from node_type& to intrusive_list_node&, I left the rest of the original code in tact.

rparolin: Yes, this is more of a question about the original code.  I'm not sure the purpose of casting the original type of the data members (which is already intrusive_list_node) to the user-type then back to its base-class.  There is case of removing const but in that case I'm not sure why we are casting to node_type.

DragoonX6: It's been a while, any updates?

rparolin:
Sorry for the delay.  Thanks for the PR.

PR #62, opened 2016-10-25
Update is_assignable to be well-defined on MSVC 2013+

CypherSignal:
Fixing an issue in type_traits where is_assignable<T, U> would not be fully defined in MSVC 2013+. For example,  assigning a value of a struct to a ref of a struct of the same type would previously fail.

is_assignable<T,U> et al should now well-defined by utilizing eastl::declval, which should function in MSVC 2013 and later. EASTL_TYPE_TRAIT_is_assignable_CONFORMANCE is now defined, as well.


rparolin:
Looks good to me.  Thanks for the PR!  :)

PR #69, opened 2016-11-11
Compatibility with VC9 and VC10

dierkes:
These fixes make EASTL usable with VC9 and VC10 (without C++11 support).


rparolin:
Looks good to me.  Thanks for the PR.

PR #75, opened 2016-12-02
Ignore IDE and build files.

DragoonX6:
Ignores Visual Studio .suo files, Qt Creator .user files,
and the folders generated by the CI scripts.

PR #76, opened 2016-12-02
Remove the need to update pacman before mingw appveyor builds.

DragoonX6:
This should speed up the process of the build somewhat.
Quite a significant portion of the time was needing to update gcc and clang to version that wouldn't crash during the build.
Appveyor now runs gcc 5.3 and clang 3.8, clang still needs to be installed, but that's not really slow.

Depends on my previous PR #75.


rparolin:
Awesome.  Thanks for the PR.

Issue #47, opened 2016-06-10
Including inttypes.h after eabase.h produces warnings on VS2015


JohanKotlinskiAtKing:
#include "EABase\eabase.h"
    #include <inttypes.h>

produces warnings on VS2015, update 2:

1>C:\Program Files (x86)\Windows Kits\10\Include\10.0.10240.0\ucrt\inttypes.h(109): warning C4005: 'PRId8': macro redefinition
1>  C:\Users\johan.kotlinski\Desktop\EASTL\test\packages\EABase\include\Common\EABase\eabase.h(402): note: see previous definition of 'PRId8'
1>C:\Program Files (x86)\Windows Kits\10\Include\10.0.10240.0\ucrt\inttypes.h(112): warning C4005: 'PRId64': macro redefinition
1>  C:\Users\johan.kotlinski\Desktop\EASTL\test\packages\EABase\include\Common\EABase\eabase.h(424): note: see previous definition of 'PRId64'
1>C:\Program Files (x86)\Windows Kits\10\Include\10.0.10240.0\ucrt\inttypes.h(125): warning C4005: 'PRIdPTR': macro redefinition
1>  C:\Users\johan.kotlinski\Desktop\EASTL\test\packages\EABase\include\Common\EABase\eabase.h(432): note: see previous definition of 'PRIdPTR'
1>C:\Program Files (x86)\Windows Kits\10\Include\10.0.10240.0\ucrt\inttypes.h(128): warning C4005: 'PRIi8': macro redefinition
1>  C:\Users\johan.kotlinski\Desktop\EASTL\test\packages\EABase\include\Common\EABase\eabase.h(403): note: see previous definition of 'PRIi8'
1>C:\Program Files (x86)\Windows Kits\10\Include\10.0.10240.0\ucrt\inttypes.h(131): warning C4005: 'PRIi64': macro redefinition
1>  C:\Users\johan.kotlinski\Desktop\EASTL\test\packages\EABase\include\Common\EABase\eabase.h(425): note: see previous definition of 'PRIi64'
1>C:\Program Files (x86)\Windows Kits\10\Include\10.0.10240.0\ucrt\inttypes.h(144): warning C4005: 'PRIiPTR': macro redefinition
1>  C:\Users\johan.kotlinski\Desktop\EASTL\test\packages\EABase\include\Common\EABase\eabase.h(433): note: see previous definition of 'PRIiPTR'
[...]

PR #81, opened 2017-01-09
Fix C99 inttypes.h macro redefinition warnings with VS2013 and above

zhiguangwang:
This commit fixes [issue 47](https://github.com/electronicarts/EASTL/issues/47).

According to [Visual C++ Team Blog](https://blogs.msdn.microsoft.com/vcblog/2013/07/19/c99-library-support-in-visual-studio-2013/), MSVC added C99 library support in Visual Studio 2013.

### DIFF
--- a/test/packages/EABase/include/Common/EABase/eabase.h
+++ b/test/packages/EABase/include/Common/EABase/eabase.h
@@ -114,8 +114,8 @@
 	   #define __STDC_FORMAT_MACROS
 	#endif
 	// The GCC PSP compiler defines standard int types (e.g. uint32_t) but not PRId8, etc.
-	// MSVC doesn't include an inttypes.h header.
-	#if !defined(EA_COMPILER_MSVC)
+	// MSVC added support for inttypes.h header in VS2013.
+	#if !defined(EA_COMPILER_MSVC) || EA_COMPILER_VERSION >= 1800
 		#include <inttypes.h> // PRId8, SCNd8, etc.
 	#endif
 	#if defined(_MSC_VER)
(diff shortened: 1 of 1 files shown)


test/packages/EABase/include/Common/EABase/eabase.h
```diff
@@ -114,8 +114,8 @@
 	   #define __STDC_FORMAT_MACROS
 	#endif
 	// The GCC PSP compiler defines standard int types (e.g. uint32_t) but not PRId8, etc.
-	// MSVC doesn't include an inttypes.h header.
-	#if !defined(EA_COMPILER_MSVC)
+	// MSVC added support for inttypes.h header in VS2013.
+	#if !defined(EA_COMPILER_MSVC) || EA_COMPILER_VERSION >= 1800
```
rparolin: Because 'EA_COMPILER_VERSION' is generic macro that is populated with the compilers version we need to restrict the check to MSVC.

#if !defined(EA_COMPILER_MSVC) || (defined(EA_COMPILER_MSVC) && EA_COMPILER_VERSION >= 1800)  // all other compilers and VS2013+

test/packages/EABase/include/Common/EABase/eabase.h
```diff
@@ -355,7 +355,7 @@
 	typedef double              double_t;
 #endif
 
-#if defined(EA_COMPILER_HAS_INTTYPES) && !defined(EA_COMPILER_MSVC)
+#if defined(EA_COMPILER_HAS_INTTYPES) && (!defined(EA_COMPILER_MSVC) || EA_COMPILER_VERSION >= 1800)
```
rparolin: Ditto above.

test/packages/EABase/include/Common/EABase/eabase.h
zhiguangwang: Actually `defined(EA_COMPILER_MSVC)` would be redundant, because the right-hand side of the `||` operator would be evaluated only if `EA_COMPILER_MSVC` has already been defined.

rparolin:
Yeah.  But we don't want other compilers with the same version to be
enabled.  EA_COMPILER_VERION is available for all compilers

rparolin:
Thank you.

Issue #68, opened 2016-11-07
EASTL doesn't detect non-apple clang on MacOS properly - thinks modern clang versions are pre-C++11


xaxxon:
on clang 3.9 (and probably others), I have to make sure to have

#define EA_HAVE_CPP11_INITIALIZER_LIST

before I include any eastl headers or I get redefinition of initializer_list.  Is this normal?

Searching the eastl source on github for that string, I don't see anywhere it would be set automatically...

Thank you.

rparolin:
The included EABase package is EASTL's only dependency.  It handles defining the necessary compiler and platform defines EASTL uses to configure itself.  The necessary EABase header files should be included implicitly by EASTL headers.

EA_HAVE_CPP11_INITIALIZER_LIST defined here:
https://github.com/electronicarts/EASTL/blob/e16f7712ec6f3857e769172c020441d358002d34/test/packages/EABase/include/Common/EABase/eahave.h#L461

FAQ entry on this:
https://github.com/electronicarts/EASTL/blob/e16f7712ec6f3857e769172c020441d358002d34/doc/EASTL%20FAQ.html#L780

Example of an EASTL header including EABase headers:
https://github.com/electronicarts/EASTL/blob/e16f7712ec6f3857e769172c020441d358002d34/include/EASTL/functional.h#L9

xaxxon:
in eacompilers.h

                #elif defined(EA_COMPILER_CPP11_ENABLED) && defined(__clang__) && (EA_COMPILER_VERSION >= 301) && !defined(__APPLE__) // Clang 3.1+, not including Apple's Clang.                           


is what's killing me.   I'm running 3.9 (real clang version, not stupid apple clang version number) downloaded directly from clang's download site:  http://llvm.org/releases/download.html

Any thoughts on a workaround?  Either for my code or for eacompilers.h directly?

rparolin:
Is your compiler not in at least C++11 mode? 

		#if (__cplusplus >= 201103L)    // Clang and GCC defines this like so in C++11 mode.
			#define EA_COMPILER_CPP11_ENABLED 1

xaxxon:
If I'm reading this correctly, 

Because this isn't true:

		#elif defined(EA_COMPILER_CPP11_ENABLED) && defined(__clang__) && (EA_COMPILER_VERSION >= 301) && !defined(__APPLE__) // Clang 3.1+, not including Apple's Clang.


it sets: 

[...]

PR #86, opened 2017-02-12
travis test for clang v 3.9 on osx; fix AppleClang detection

theryee:
Fix for issue #68


test/packages/EABase/include/Common/EABase/config/eacompiler.h
```diff
@@ -726,9 +726,9 @@
 			// Extern template is supported.
 		#elif defined(EA_COMPILER_CPP11_ENABLED) && defined(__EDG_VERSION__) && (__EDG_VERSION__ >= 401)  // EDG 4.1+.
 			// supported.
-		#elif defined(EA_COMPILER_CPP11_ENABLED) && defined(__clang__) && defined(__APPLE__) && (EA_COMPILER_VERSION >= 401)
+		#elif defined(EA_COMPILER_CPP11_ENABLED) && defined(__clang__) && defined(__apple_build_version__) && (EA_COMPILER_VERSION >= 401)
```
rparolin: Why is the change from '__APPLE__' to '__apple_build_version__' required?

theryee: __APPLE__ identifies the platform (Max OS X)
__apple_build_version__ identifies the compiler (AppleClang)

https://sourceforge.net/p/predef/wiki/Compilers/
https://sourceforge.net/p/predef/wiki/Home/

rparolin: Yup.  Got that.  But we are already testing for both "__APPLE__ && __clang__" which would be equivalent.  Unless I'm missing something?  The __clang__ would already prevent using GCC on MacOSX, for example.

theryee: Ah you know what I just did a s/__APPLE__/__apple_build_version__/ on this file because I saw a couple bad examples. Let me revisit... In hindsight, I think the combination of __APPLE__ __clang__ is clearer than simply __apple_build_version__.

theryee: Ahh I remember the problem now... if I have clang3.9 running on mac OS X, this is true:
defined(__clang__) && defined(__APPLE__)
because it is the case that the compiler is clang and the platform is apple.

rparolin: Why is that a problem?  Sorry, I'm not following the issue that was causing as we change the "EA_COMPILER_VERSION" checks to use the AppleClang version numbers.

theryee: The problem is clang3.9 is not AppleClang. It's just plain old "Clang" and so 3.9 looks like a really old version of AppleClang because 3.9 < 4.0.1 but really it should be hit the non-apple clang comparison (where any version of "Clang" is ok).

See this snippet:
		#elif defined(EA_COMPILER_CPP11_ENABLED) && defined(__clang__) && defined(__apple_build_version__) && (EA_COMPILER_VERSION >= 401)
			// Extern template is supported.
		#elif defined(EA_COMPILER_CPP11_ENABLED) && defined(__clang__) && !defined(__apple_build_version__)             // Clang other than Apple's Clang
			// Extern template is supported.

rparolin:
Awesome!  Thanks for the explanation. :)

Issue #85, opened 2017-02-09
No at() in eastl::map


bnason-nf:
Hi,

It looks like C++ 11 now supports an at() function in [map](http://en.cppreference.com/w/cpp/container/map/at) I guess this should probably be added to the eastl map API?

Thanks,
Benbuck

PR #87, opened 2017-02-14
mapAt

theryee:
Issue #85

### DIFF
--- a/include/EASTL/map.h
+++ b/include/EASTL/map.h
@@ -447,17 +447,14 @@ namespace eastl
 	{
 		iterator itLower(lower_bound(key)); // itLower->first is >= key.
 
-		#if EASTL_EXCEPTIONS_ENABLED
-			if(itLower == end())
-			{
+		if(itLower == end())
+		{
+			#if EASTL_EXCEPTIONS_ENABLED
 				throw std::out_of_range("map::at key does not exist");
-			}
-		#else
-			if(itLower == end())
-			{
+			#else
 				EASTL_FAIL_MSG("map::at key does not exist");
-			}
-		#endif
+			#endif
+		}
 
 		return (*itLower).second;
 	}
@@ -468,17 +465,14 @@ namespace eastl
 	{
 		const_iterator itLower(lower_bound(key)); // itLower->first is >= key.
 
-		#if EASTL_EXCEPTIONS_ENABLED
-			if(itLower == end())
-			{
+		if(itLower == end())
+		{
+			#if EASTL_EXCEPTIONS_ENABLED
 				throw std::out_of_range("map::at key does not exist");
-			}
-		#else
-			if(itLower == end())
-			{
+			#else
 				EASTL_FAIL_MSG("map::at key does not exist");
-			}
-		#endif
+			#endif
+		}
 
 		return (*itLower).second;
 	}
--- a/include/EASTL/map.h
+++ b/include/EASTL/map.h
@@ -11,6 +11,7 @@
 #include <EASTL/internal/red_black_tree.h>
 #include <EASTL/functional.h>
 #include <EASTL/utility.h>
+#include <stdexcept>
 
 #if defined(EA_PRAGMA_ONCE_SUPPORTED)
 	#pragma once // Some compilers (e.g. VC++) benefit significantly from using this. We've measured 3-4% build speed improvements in apps as a result.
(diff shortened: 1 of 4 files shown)


theryee:
Does it make sense to add at() to other map-like structures in the EASTL?

include/EASTL/map.h
```diff
@@ -11,6 +11,7 @@
 #include <EASTL/internal/red_black_tree.h>
 #include <EASTL/functional.h>
 #include <EASTL/utility.h>
+#include <stdexcept>
```
rparolin: This shouldn't be necessary as it should be added in the config.h header.

include/EASTL/map.h
```diff
@@ -439,6 +443,33 @@ namespace eastl
...
 
 
+	template <typename Key, typename T, typename Compare, typename Allocator>
+	inline T& map<Key, T, Compare, Allocator>::at(const Key& key)
+	{
+		iterator itLower(lower_bound(key)); // itLower->first is >= key.
+
+		if(itLower == end())
+		{
+			throw std::out_of_range("map::at key does not exist");
```
rparolin: EASTL can't assume that exceptions are enabled.  The vast majority of the time we disable exceptions so we typical add asserts in the case when exceptions are disabled.  You can see examples of how this done through the library but most recently I implemented string_view with similar code.

https://github.com/electronicarts/EASTL/blob/master/include/EASTL/string_view.h#L89

test/packages/EATest/include/EATest/EATest.h
```diff
@@ -341,6 +341,17 @@ namespace EA
             #define EATEST_VERIFY_F(bExpression, pFormat, ...) EA::UnitTest::TestInternal::EATEST_VERIFY_F_IMP((bExpression), nErrorCount, __FILE__, __LINE__, pFormat, __VA_ARGS__)
         #endif
 
+				/// EATEST_VERIFY_THROW
+				/// EATEST_VERIFY_NOTHROW
+				/// 
+				/// This macro confirms whether or not an expression throws or doesn't throw.
+				/// 
+				/// See EATEST_VERIFY for details about error reporting and the _MSG variants
+				#define EATEST_VERIFY_THROW(expression) {bool isNoThrow; try{ {expression;} isNoThrow=true; }catch(...){isNoThrow=false;} EATEST_VERIFY(!isNoThrow); }
```
rparolin: Similarly code that assumes to throw must be wrapped with exception preprocessor guards.

test/source/TestMap.cpp
```diff
@@ -141,6 +140,23 @@ int TestMap()
 		EATEST_VERIFY(v.validate());
 	}
 
+	{
+		typedef eastl::map<int, int>     IntIntMap;
+		IntIntMap map1;
+
+		EATEST_VERIFY_THROW(map1.at(0));
+		map1[0]=1;
+		EATEST_VERIFY_NOTHROW(map1.at(0));
```
rparolin: Again preprocessor guards for exceptions.

include/EASTL/map.h
theryee: I'll trace this out. I was getting missing declaration errors on linux gcc.

include/EASTL/map.h
theryee: k

test/packages/EATest/include/EATest/EATest.h
theryee: k

test/source/TestMap.cpp
theryee: k

include/EASTL/map.h
```diff
@@ -439,6 +442,47 @@ namespace eastl
 	#endif
 
 
+	template <typename Key, typename T, typename Compare, typename Allocator>
+	inline T& map<Key, T, Compare, Allocator>::at(const Key& key)
+	{
+		iterator itLower(lower_bound(key)); // itLower->first is >= key.
+
+		#if EASTL_EXCEPTIONS_ENABLED
+			if(itLower == end())
```
rparolin: Since the branch is checked in both cases, I'd promote it outside of the preprocessor guards.

rparolin: Other than that, I think its ready to go.

rparolin:
Awesome!  Thanks for the PR.

Issue #82, opened 2017-01-27
Can't use move semantics with hash_map::operator[]


bnason-nf:
This fails to compile. Is this a problem with EASTL or in my understanding of how the hash_map can be used? I am using 3.05.00.

Thanks,
Benbuck

```c++
#include "eastl/hash_map.h"

class Key
{
public:
    Key() {}
    Key(Key && o) {}
    bool operator==(const Key& other) const { return true; }
private:
    Key(const Key & o) {}
};

struct Hash
{
    std::size_t operator()(const Key& k) const { return 0; }
};

int test(void)
{
    Key key;
    eastl::hash_map<Key, int, Hash> hm;
    return hm[eastl::move(key)];
}
```
The error emitted by my compiler is:

```
eastl\include\EASTL\utility.h(401,10): error 331-D: "Key::Key(const Key &)" (declared at line 10 of "test.cpp") is inaccessible
            detected during:
              eastl\include\EASTL\utility.h(47491,48): instantiation of "eastl::pair<T1, T2>::pair(const T1 &) [with T1=const Key, T2=int]" at line 1442 of "eastl\include\EASTL\internal\hashtable.h"
              eastl\include\EASTL\utility.h(48466,52): instantiation of "eastl::hashtable<Key, Value, Allocator, ExtractKey, Equal, H1, H2, H, RehashPolicy, bCacheHashCode, bMutableIterators, bUniqueKeys>::node_type *eastl::hashtable<Key, Value, Allocator, ExtractKey, Equal, H1, H2, H, RehashPolicy, bCacheHashCode, bMutableIterators, bUniqueKeys>::DoAllocateNodeFromKey(const eastl::hashtable<Key, Value, Allocator, ExtractKey, Equal, H1, H2, H, RehashPolicy, bCacheHashCode, bMutableIterators, bUniqueKeys>::key_type &) [with Key=Key, Value=eastl::pair<const Key, int>, Allocator=eastl::allocator, ExtractKey=eastl::use_first<eastl::pair<const Key, int>>, Equal=eastl::equal_to<Key>, H1=Hash, H2=eastl::mod_range_hashing, H=eastl::default_ranged_hash, RehashPolicy=eastl::prime_rehash_policy, bCacheHashCode=false, bMutableIterators=true, bUniqueKeys=true]" at line 2417 of "eastl\include\EASTL\internal\hashtable.h"
[...]

PR #95, opened 2017-03-25
Fix for issue #82; move semantics with hash_map::operator[]

theryee:


test/source/TestUtility.cpp
```diff
@@ -543,6 +554,20 @@ static int TestUtilityExchange()
...
+	#if EASTL_MOVE_SEMANTICS_ENABLED
+	{
+		int i1 = 1;
+		TestPairSingleMoveConstructor<int> test1;
+		test1.test(eastl::move(i1));
+
+		int i2a = 2;
+		int & i2b = i2a;
+		TestPairSingleMoveConstructor<int&> test2;
+		test1.test(eastl::move(i2b));
```
rparolin: shouldn't this be test2.test(...)?    I'd also add scopes around the test cases here to avoid name scoping issues like that in the future.

test/source/TestHash.cpp
```diff
@@ -1274,6 +1274,41 @@ int TestHash()
...
+		};
+
+		struct Hash
+		{
+			std::size_t operator()(const Key& k) const { return 0; }
+		};
+
+		class Tester
+		{
+			public:
```
theryee: K

test/source/TestUtility.cpp
theryee: oops! Yeah I'll fix that.

theryee: I need to massage the meta functions for this constructor. test2 actually fails to compile.

theryee: Ok I think I have a new candidate:
```
			template <typename U, typename = typename eastl::enable_if<eastl::is_convertible<U&&,typename eastl::remove_reference<T1>::type&&>::value>::type>
			EA_CPP14_CONSTEXPR pair(U&& x)
				: first(eastl::move(x)),
					second()
			{
			}
```

The problem is I don't think this is a useful test:
`test2.test(eastl::move(i2b))`

Because move basically does this:
`static_cast<typename eastl::remove_reference<decltype(x)>::type&&>(x)`

The critical part being the remove_reference<> applied to reference generates a temporary int... which results in this error:
```
EASTL/test/../include/EASTL/utility.h:420:7: error: non-const lvalue reference to type 'int' cannot bind to a temporary of type 'typename eastl::remove_reference<int &>::type'
      (aka 'int')
                                : first(eastl::move(x)),
                                  ^     ~~~~~~~~~~~~~~
EASTL/test/source/TestUtility.cpp:31:24: note: in instantiation of function template specialization 'eastl::pair<int &, int>::pair<int, void>' requested here
                                eastl::pair<T1,T2> p(eastl::move(t1));
                                                   ^
EASTL/test/source/TestUtility.cpp:578:10: note: in instantiation of member function 'TestPairSingleMoveConstructor<int &, int>::test2' requested here
                        test2.test2(i2b);
                              ^
```

rparolin: I'd agree that this isn't a useful test. What does it mean to "move" an object  into a reference.  If it even compiled, it would have to create temporaries for the reference to bind to which would be terminated leaving a dangling reference.

I think its fine to remove that test.

theryee:
Yeah my thoughts exactly. Creating a rvalue ref for a ref to a temp is not a practical use case. I'll remove this test and experiment with creating a template to remove the code duplication that was noted.

rparolin:
Awesome.  Thank you.

theryee:
Couldn't quite get the type switching for "const T &"/"T&&" to work... couldn't get the overload to select the move type instead of the reference type.

rparolin:
> Couldn't quite get the type switching for "const T &"/"T&&" to work... couldn't get the overload to select the move type instead of the reference type.

Sorry, I'm not following.

theryee:
I tried doing something like this:

```
mapped_type& operator[](key_type&& key){
return (*base_type::template DoInsertKey<key_type&&>(true_type(), eastl::move(key)).first).second;
}

template <typename T, typename = typename eastl::enable_if<eastl::is_convertible<T,key_type>::value>::type>
node_type*  DoAllocateNodeFromKey(T key);

template<typename T, typename = typename eastl::enable_if<eastl::is_convertible<T,key_type>::value>::type>
eastl::pair<iterator, bool> DoInsertKey(true_type, T key);
```

But it was still selecting the copy constructor and not the move constructor.

Furthermore, even if the correct T is injected, we still need inject a east::move:
```
node_type* const pNodeNew = DoAllocateNodeFromKey(eastl::move(key));
```

So I think without significant refactoring to decouple the specific DoAllocateNodeFromKey call from DoInsertKey... I think reducing the code duplication is a fairly invasive change.

test/source/TestUtility.cpp
```diff
@@ -543,6 +543,23 @@ static int TestUtilityExchange()
 		EATEST_VERIFY(a == "EA Standard Template Library");
 	}
 
+	// Issue #82; construct pair using single move constructor
+	#if EASTL_MOVE_SEMANTICS_ENABLED
+	{
+		class TestPairSingleMoveConstructor
```
rparolin: Fix the intend formatting here.  Since you are changing might as well make it a struct and remove the public.  Left justify the '&&'.  You can can also just use clang-format as EASTL has its format script in the package root.

theryee: K

rparolin:
Thanks for the PR!

Issue #92, opened 2017-02-17
Unable to sort non-copy-constructible objects


sekmet64:
The code below fails to compile with eastl but works fine with libc++
```
std::vector<std::unique_ptr<int>> vec;
eastl::sort(vec.begin(), vec.end());
```

rparolin:
Thanks for submitting this.  I take a look at it.

rparolin:
Is this an std/eastl interop bug?  Or are you saying eastl::vector<eastl::unique_ptr>>  fails to compile when passd to the eastl::sort?

sekmet64:
No sorry for being too vague, it's the lack of support for only move-constructible objects in the sorting algorithms. The same error can be reproduced on both msvc and clang with:

```
eastl::vector<eastl::unique_ptr<int>> vec;
eastl::sort(vec.begin(), vec.end(),  [](const eastl::unique_ptr<int>& lhs, const eastl::unique_ptr<int>& rhs) { return *lhs < *rhs; });
```

theryee:
Been experimenting with this one. Fix for this looks fairly involved as I believe it involves changing sort.h's move support.

rparolin:
Yes, it is an involved change.  I have this work tracked in our internal backlog but so I will get around to doing it at some point in the future.

theryee:
I think I got this figured out... just need to fix some build issues for some of the compilers.

theryee:
I've posted an updated review.

rparolin:
Sound be resolved with release 3.05.06.

theryee:
I think this is still an issue? I tried applying only the most basic test here:
https://github.com/theryee/EASTL/commit/7623f1023718330129a3604c570a39752d3c5291

But still see compile errors regarding attempting to use the deleted copy constructor for unique_ptr:
https://travis-ci.org/theryee/EASTL/builds/239803490?utm_source=github_status&utm_medium=notification

theryee:
Note: my PR is still open and un-merged... did you implement a fix independently?

PR #101, opened 2017-04-23
Unable to sort non-copy-constructible objects (Issue #92)

theryee:
Issue #92

### DIFF
--- a/include/EASTL/algorithm.h
+++ b/include/EASTL/algorithm.h
@@ -967,6 +967,33 @@ namespace eastl
 		return b;
 	}
 
+	/// median
+	///
+	/// median finds which element of three (a, b, d) is in-between the other two.
+	/// If two or more elements are equal, the first (e.g. a before b) is chosen.
+	///
+	/// Complexity: Either two or three comparisons will be required, depending 
+	/// on the values.
+	///
+	template <typename T>
+	inline T&& median(T&& a, T&& b, T&& c)
+	{
+		if(a < b)
+		{
+			if(b < c)
+				return eastl::move(b);
+			else if(a < c)
+				return eastl::move(c);
+			else
+				return eastl::move(a);
+		}
+		else if(a < c)
+			return eastl::move(a);
+		else if(b < c)
+			return eastl::move(c);
+		return eastl::move(b);
+	}
+
 
 	/// median
 	///
@@ -993,6 +1020,33 @@ namespace eastl
 		else if(compare(b, c))
 			return c;
 		return b;
+	}
+
+	/// median
+	///
+	/// median finds which element of three (a, b, d) is in-between the other two.
+	/// If two or more elements are equal, the first (e.g. a before b) is chosen.
+	///
+	/// Complexity: Either two or three comparisons will be required, depending 
+	/// on the values.
+	///
+	template <typename T, typename Compare>
+	inline T&& median(T&& a, T&& b, T&& c, Compare compare)
+	{
+		if(compare(a, b))
+		{
+			if(compare(b, c))
+				return eastl::move(b);
+			else if(compare(a, c))
+				return eastl::move(c);
+			else
+				return eastl::move(a);
+		}
+		else if(compare(a, c))
+			return eastl::move(a);
+		else if(compare(b, c))
+			return eastl::move(c);
+		return eastl::move(b);
 	}
 
 
--- a/include/EASTL/algorithm.h
+++ b/include/EASTL/algorithm.h
@@ -940,7 +940,7 @@ namespace eastl
 	}
 
 	template <typename T>
-	inline T median_impl(T a, T b, T c)
+	inline T&& median_impl(T&& a, T&& b, T&& c)
 	{
 		if(a < b)
 		{
@@ -969,7 +969,7 @@ namespace eastl
 	template <typename T>
 	inline const T& median(const T& a, const T& b, const T& c)
 	{
-		return median_impl<const T&>(a, b, c);
+		return median_impl(a, b, c);
 	}
 
 	/// median
@@ -983,12 +983,12 @@ namespace eastl
 	template <typename T>
 	inline T&& median(T&& a, T&& b, T&& c)
 	{
-		return eastl::forward<T>(median_impl<T&&>(eastl::forward<T>(a), eastl::forward<T>(b), eastl::forward<T>(c)));
+		return eastl::forward<T>(median_impl(eastl::forward<T>(a), eastl::forward<T>(b), eastl::forward<T>(c)));
 	}
 
 
 	template <typename T, typename Compare>
-	inline T median_impl(T a, T b, T c, Compare compare)
+	inline T&& median_impl(T&& a, T&& b, T&& c, Compare compare)
 	{
 		if(compare(a, b))
 		{
--- a/include/EASTL/algorithm.h
+++ b/include/EASTL/algorithm.h
@@ -939,6 +939,24 @@ namespace eastl
 		return eastl::make_pair(*iteratorPair.first, *iteratorPair.second);
 	}
 
+	template <typename T>
+	inline T median_impl(T a, T b, T c)
+	{
+		if(a < b)
+		{
+			if(b < c)
+				return eastl::forward<T>(b);
+			else if(a < c)
+				return eastl::forward<T>(c);
+			else
+				return eastl::forward<T>(a);
+		}
+		else if(a < c)
+			return eastl::forward<T>(a);
+		else if(b < c)
+			return eastl::forward<T>(c);
+		return eastl::forward<T>(b);
+	}
 
 	/// median
 	///
@@ -951,20 +969,7 @@ namespace eastl
 	template <typename T>
 	inline const T& median(const T& a, const T& b, const T& c)
 	{
-		if(a < b)
-		{
-			if(b < c)
-				return b;
-			else if(a < c)
-				return c;
-			else
-				return a;
-		}
-		else if(a < c)
-			return a;
-		else if(b < c)
-			return c;
-		return b;
+		return median_impl<const T&>(a, b, c);
 	}
 
 	/// median
@@ -978,20 +983,27 @@ namespace eastl
 	template <typename T>
 	inline T&& median(T&& a, T&& b, T&& c)
 	{
-		if(a < b)
+		return eastl::forward<T>(median_impl<T&&>(eastl::forward<T>(a), eastl::forward<T>(b), eastl::forward<T>(c)));
+	}
+
+
+	template <typename T, typename Compare>
+	inline T median_impl(T a, T b, T c, Compare compare)
+	{
+		if(compare(a, b))
 		{
-			if(b < c)
-				return eastl::move(b);
-			else if(a < c)
-				return eastl::move(c);
+			if(compare(b, c))
+				return eastl::forward<T>(b);
+			else if(compare(a, c))
+				return eastl::forward<T>(c);
 			else
-				return eastl::move(a);
+				return eastl::forward<T>(a);
 		}
-		else if(a < c)
-			return eastl::move(a);
-		else if(b < c)
-			return eastl::move(c);
-		return eastl::move(b);
+		else if(compare(a, c))
+			return eastl::forward<T>(a);
+		else if(compare(b, c))
+			return eastl::forward<T>(c);
+		return eastl::forward<T>(b);
 	}
 
 
@@ -1006,20 +1018,7 @@ namespace eastl
 	template <typename T, typename Compare>
 	inline const T& median(const T& a, const T& b, const T& c, Compare compare)
 	{
-		if(compare(a, b))
-		{
-			if(compare(b, c))
-				return b;
-			else if(compare(a, c))
-				return c;
-			else
-				return a;
-		}
-		else if(compare(a, c))
-			return a;
-		else if(compare(b, c))
-			return c;
-		return b;
+		return median_impl<const T&, Compare>(a, b, c, compare);
 	}
 
 	/// median
@@ -1033,20 +1032,7 @@ namespace eastl
 	template <typename T, typename Compare>
 	inline T&& median(T&& a, T&& b, T&& c, Compare compare)
 	{
-		if(compare(a, b))
-		{
-			if(compare(b, c))
-				return eastl::move(b);
-			else if(compare(a, c))
-				return eastl::move(c);
-			else
-				return eastl::move(a);
-		}
-		else if(compare(a, c))
-			return eastl::move(a);
-		else if(compare(b, c))
-			return eastl::move(c);
-		return eastl::move(b);
+		return eastl::forward<T>(median_impl<T&&, Compare>(eastl::forward<T>(a), eastl::forward<T>(b), eastl::forward<T>(c), compare));
 	}
 
 
--- a/include/EASTL/heap.h
+++ b/include/EASTL/heap.h
@@ -213,7 +213,6 @@ namespace eastl
 	/// This function requires that the value argument refer to a value
 	/// that is currently not within the heap.
 	///
-	/// TODO: reduce code duplication here
 	template <typename RandomAccessIterator, typename Distance, typename T>
 	void adjust_heap(RandomAccessIterator first, Distance topPosition, Distance heapSize, Distance position, T&& value)
 	{
--- a/include/EASTL/heap.h
+++ b/include/EASTL/heap.h
@@ -62,8 +62,7 @@ namespace eastl
 	/// This function requires that the value argument refer to a value
 	/// that is currently not within the heap.
 	///
-	template <typename RandomAccessIterator, typename Distance, typename T,
-		typename = typename eastl::enable_if<eastl::is_copy_assignable<typename eastl::iterator_traits<RandomAccessIterator>::value_type>::value>::type>
+	template <typename RandomAccessIterator, typename Distance, typename T>
 	inline void promote_heap(RandomAccessIterator first, Distance topPosition, Distance position, const T& value)
 	{
 		for(Distance parentPosition = (position - 1) >> 1; // This formula assumes that (position > 0). // We use '>> 1' instead of '/ 2' because we have seen VC++ generate better code with >>.
@@ -89,19 +88,20 @@ namespace eastl
 	/// This function requires that the value argument refer to a value
 	/// that is currently not within the heap.
 	///
-	template <typename RandomAccessIterator, typename Distance, typename T,
-		typename = typename eastl::enable_if<eastl::is_move_assignable<typename eastl::iterator_traits<RandomAccessIterator>::value_type>::value>::type>
+	template <typename RandomAccessIterator, typename Distance, typename T>
 	inline void promote_heap(RandomAccessIterator first, Distance topPosition, Distance position, T&& value)
 	{
+		typedef typename iterator_traits<RandomAccessIterator>::value_type value_type;
+
 		for(Distance parentPosition = (position - 1) >> 1; // This formula assumes that (position > 0). // We use '>> 1' instead of '/ 2' because we have seen VC++ generate better code with >>.
 			(position > topPosition) && (*(first + parentPosition) < value);
 			parentPosition = (position - 1) >> 1)
 		{
-			*(first + position) = eastl::move(*(first + parentPosition)); // Swap the node with its parent.
+			*(first + position) = eastl::forward<value_type>(*(first + parentPosition)); // Swap the node with its parent.
 			position = parentPosition;
 		}
 
-		*(first + position) = eastl::move(value);
+		*(first + position) = eastl::forward<value_type>(value);
 	}
 
 
@@ -116,8 +116,7 @@ namespace eastl
 	/// This function requires that the value argument refer to a value
 	/// that is currently not within the heap.
 	///
-	template <typename RandomAccessIterator, typename Distance, typename T, typename Compare,
-		typename = typename eastl::enable_if<eastl::is_copy_assignable<typename eastl::iterator_traits<RandomAccessIterator>::value_type>::value>::type>
+	template <typename RandomAccessIterator, typename Distance, typename T, typename Compare>
 	inline void promote_heap(RandomAccessIterator first, Distance topPosition, Distance position, const T& value, Compare compare)
 	{
 		for(Distance parentPosition = (position - 1) >> 1; // This formula assumes that (position > 0). // We use '>> 1' instead of '/ 2' because we have seen VC++ generate better code with >>.
@@ -143,19 +142,20 @@ namespace eastl
 	/// This function requires that the value argument refer to a value
 	/// that is currently not within the heap.
 	///
-	template <typename RandomAccessIterator, typename Distance, typename T, typename Compare,
-		typename = typename eastl::enable_if<eastl::is_move_assignable<typename eastl::iterator_traits<RandomAccessIterator>::value_type>::value>::type>
+	template <typename RandomAccessIterator, typename Distance, typename T, typename Compare>
 	inline void promote_heap(RandomAccessIterator first, Distance topPosition, Distance position, T&& value, Compare compare)
 	{
+		typedef typename iterator_traits<RandomAccessIterator>::value_type value_type;
+		
 		for(Distance parentPosition = (position - 1) >> 1; // This formula assumes that (position > 0). // We use '>> 1' instead of '/ 2' because we have seen VC++ generate better code with >>.
 			(position > topPosition) && compare(*(first + parentPosition), value);
 			parentPosition = (position - 1) >> 1)
 		{
-			*(first + position) = eastl::move(*(first + parentPosition)); // Swap the node with its parent.
+			*(first + position) = eastl::forward<value_type>(*(first + parentPosition)); // Swap the node with its parent.
 			position = parentPosition;
 		}
 
-		*(first + position) = eastl::move(value);
+		*(first + position) = eastl::forward<value_type>(value);
 	}
 
 
@@ -212,6 +212,8 @@ namespace eastl
 	template <typename RandomAccessIterator, typename Distance, typename T>
 	void adjust_heap(RandomAccessIterator first, Distance topPosition, Distance heapSize, Distance position, T&& value)
 	{
+		typedef typename iterator_traits<RandomAccessIterator>::value_type value_type;
+
 		// We do the conventional approach of moving the position down to the 
 		// bottom then inserting the value at the back and moving it up.
 		Distance childPosition = (2 * position) + 2;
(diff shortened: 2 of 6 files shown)


test/source/TestAlgorithm.cpp
```diff
@@ -2339,6 +2341,48 @@ int TestAlgorithm()
 		}
 	}
 
+	// issue #92
+	// disable in MSVC2013 because eastl::is_copy_constructible is always true...
```
theryee: I gave up on trying to work around this bug and just disabled the test for MSVC2013.

theryee: I gave up trying to find a workaround and instead disabled this test for MSVC 2013.

rparolin:
I only run clang format on the code that I'm editing.  Please revert the
formatting changes as it creates diffs unrelated to the functionality
changes.

Thanks.

theryee:
Sure that makes sense. I dropped that commit.

rparolin:
Cheers.  Thank you.

include/EASTL/sort.h
```diff
@@ -738,15 +813,39 @@ namespace eastl
 		// This is because the implementation below makes an assumption about the input
 		// data that quick_sort satisfies but arbitrary data may not.
 		// There is a standalone insertion_sort function. 
-		template <typename RandomAccessIterator, typename Compare>
+		template <typename RandomAccessIterator, typename Compare, typename eastl::enable_if<!eastl::is_lvalue_reference<typename eastl::iterator_traits<RandomAccessIterator>::value_type>::value>::type* = nullptr >
```
rparolin: Sorry for the delay in review.  

Don't want to check if the type has a move-constructor instead of if its an lvalue reference?

include/EASTL/sort.h
```diff
@@ -707,6 +754,34 @@ namespace eastl
 	}
 
 
+	template <typename RandomAccessIterator, typename T, typename Compare>
+	inline RandomAccessIterator get_partition(RandomAccessIterator first, RandomAccessIterator last, const T&& pivotValue, Compare compare, typename eastl::enable_if<eastl::is_move_constructible<T>::value>::type* = 0 )
```
rparolin: Unfortunately, Github review tools don't make horizontal scrolling easy.  Please try to stay within a 120 column width.  I'd recommend using clang-format with our provided config file.

theryee: Need to upgrade my install of vim... it seems to be too old for clang-format.py :(

include/EASTL/sort.h
theryee: Yeah that was a typo.

theryee: Oh not a typo-- re-read your comment and what I wrote... I don't remember why I used lvalue/rvalue for this case but it looks wrong. I changed it is_move_constructible instead.

include/EASTL/heap.h
```diff
@@ -275,20 +358,42 @@ namespace eastl
...
+	///
+	//template <typename RandomAccessIterator, typename Compare>
+	template <typename RandomAccessIterator, typename Compare, typename = typename eastl::enable_if<eastl::is_copy_constructible<typename eastl::iterator_traits<RandomAccessIterator>::value_type>::value>::type>
 	inline void pop_heap(RandomAccessIterator first, RandomAccessIterator last, Compare compare)
 	{
 		typedef typename eastl::iterator_traits<RandomAccessIterator>::difference_type difference_type;
 		typedef typename eastl::iterator_traits<RandomAccessIterator>::value_type      value_type;
 
-		const value_type tempBottom(*(last - 1));
+		const value_type tempBottom( *(last - 1) );
```
rparolin: no extra within parens.

include/EASTL/heap.h
```diff
@@ -275,20 +358,42 @@ namespace eastl
 	/// to make and maintain the heap.
 	///
 	template <typename RandomAccessIterator, typename Compare>
+	inline void pop_heap(RandomAccessIterator first, RandomAccessIterator last, Compare compare, typename eastl::enable_if<eastl::is_move_constructible<typename eastl::iterator_traits<RandomAccessIterator>::value_type>::value && !eastl::is_copy_constructible<typename eastl::iterator_traits<RandomAccessIterator>::value_type>::value>::type* = 0)
+	{
+		typedef typename eastl::iterator_traits<RandomAccessIterator>::difference_type difference_type;
+		typedef typename eastl::iterator_traits<RandomAccessIterator>::value_type      value_type;
+
+		value_type tempBottom( eastl::move(*(last - 1)) );
+		*(last - 1) = eastl::move( *first );
```
rparolin: no extra spaces within parens

include/EASTL/heap.h
```diff
@@ -103,6 +103,33 @@ namespace eastl
...
+	/// Takes a Compare(a, b) function (or function object) which returns true if a < b.
+	/// For example, you could use the standard 'less' comparison object.
+	///
+	/// The Compare function must work equivalently to the compare function used
+	/// to make and maintain the heap.
+	///
+	/// This function requires that the value argument refer to a value
+	/// that is currently not within the heap.
+	///
+	template <typename RandomAccessIterator, typename Distance, typename T, typename Compare, typename = typename eastl::enable_if<eastl::is_move_constructible<typename eastl::iterator_traits<RandomAccessIterator>::value_type>::value>::type>
```
rparolin: since the function body is using move-assignment, we should use the eastl::is_move_assignable type trait for the sfinae expression.  or you could get a compiler error if this function is used on a type that has a move ctor but no move assignment implemented.

include/EASTL/heap.h
```diff
@@ -87,7 +87,7 @@ namespace eastl
 	/// This function requires that the value argument refer to a value
 	/// that is currently not within the heap.
 	///
-	template <typename RandomAccessIterator, typename Distance, typename T, typename Compare>
+	template <typename RandomAccessIterator, typename Distance, typename T, typename Compare, typename = typename eastl::enable_if<eastl::is_copy_constructible<typename eastl::iterator_traits<RandomAccessIterator>::value_type>::value>::type>
```
rparolin: similar issue here... this is function body using copy-assignment.

also, please add unit tests to validate this behaviour.

include/EASTL/sort.h
```diff
@@ -213,6 +213,54 @@ namespace eastl
...
+	{
+		typedef typename eastl::iterator_traits<BidirectionalIterator>::value_type value_type;
+
+		if(first != last) // if the range is non-empty...
+		{
+			BidirectionalIterator iCurrent, iNext, iSorted = first;
+
+			for(++iSorted; iSorted != last; ++iSorted)
+			{
+				value_type temp( eastl::move(*iSorted) );
```
rparolin: no extra space within parens.

include/EASTL/heap.h
rparolin: why do we have the extra requirement of T not being copy-constructible here?

include/EASTL/heap.h
```diff
@@ -324,6 +429,28 @@ namespace eastl
 
 
 	template <typename RandomAccessIterator, typename Compare>
+	void make_heap(RandomAccessIterator first, RandomAccessIterator last, Compare compare, typename eastl::enable_if<eastl::is_move_constructible<typename eastl::iterator_traits<RandomAccessIterator>::value_type>::value && !eastl::is_copy_constructible<typename eastl::iterator_traits<RandomAccessIterator>::value_type>::value>::type* = 0)
```
rparolin: why do we have the extra requirement of T not being copy_constructible?

include/EASTL/sort.h
```diff
@@ -213,6 +213,54 @@ namespace eastl
 	/// inner for loop with a call to upper_bound, which would be faster.
 	///
 	template <typename BidirectionalIterator, typename StrictWeakOrdering>
+	void insertion_sort(
+	    BidirectionalIterator first,
+	    BidirectionalIterator last,
+	    StrictWeakOrdering compare,
+	    typename eastl::enable_if<
+	        eastl::is_move_constructible<typename eastl::iterator_traits<BidirectionalIterator>::value_type>::value>::
```
rparolin: check for is_move_assignable as well... as the  function body uses move assignment as well.

include/EASTL/sort.h
```diff
@@ -213,6 +213,54 @@ namespace eastl
...
+	/// a for loop. If we were to specialize this for a RandomAccessIterator, we could replace the
+	/// inner for loop with a call to upper_bound, which would be faster.
+	///
+	template <typename BidirectionalIterator,
+	          typename StrictWeakOrdering,
+	          typename = typename eastl::enable_if<!eastl::is_move_constructible<
+	              typename eastl::iterator_traits<BidirectionalIterator>::value_type>::value>::type>
 	void insertion_sort(BidirectionalIterator first, BidirectionalIterator last, StrictWeakOrdering compare)
 	{
 		typedef typename eastl::iterator_traits<BidirectionalIterator>::value_type value_type;
```
rparolin: Actually, do we even need these additional overloads that we sfinae?  I believe we can just add the eastl::move to the original implementations and if the types aren't movable it should degrade to a copy behaviour.  That would simplify the implementation we could do this and would remove alot of duplicated code.

test/source/TestAlgorithm.cpp
```diff
@@ -2339,6 +2341,48 @@ int TestAlgorithm()
...
+			vec.emplace_back(new int(7));
+			vec.emplace_back(new int(-42));
+			vec.emplace_back(new int(5));
+			eastl::sort(vec.begin(), vec.end(),  [](const eastl::unique_ptr<int>& lhs, const eastl::unique_ptr<int>& rhs) { return *lhs < *rhs; });
+			EATEST_VERIFY(*vec[0] == -42);
+			EATEST_VERIFY(*vec[1] == 5);
+			EATEST_VERIFY(*vec[2] == 7);
+		}
+		{
+			for(unsigned tests=0; tests<500; ++tests)
```
rparolin: why is this test being run 500 times?

include/EASTL/sort.h
theryee: bad habit of mine will remove.

include/EASTL/sort.h
theryee: yeah makes sense.

include/EASTL/heap.h
theryee: ok

include/EASTL/heap.h
theryee: Fixed the formatting...

I think I must've painted myself into a corner with a previous version of this change... Tested this with just using is_move_constructible and it compiles and passes. Will fix.

include/EASTL/heap.h
theryee: OK, will change and add test.

include/EASTL/heap.h
theryee: OK

include/EASTL/heap.h
theryee: Fixed

test/source/TestAlgorithm.cpp
theryee: I was imitating the "strategy" implemented by random sorting test at line 2177. The principal appeared to be:
 * sort a small array
 * do the sorts many times to gain more confidence.

include/EASTL/sort.h
theryee: Let me try that out. That would certainly reduce code duplication. My reading of how it works with with respect to casting and rvalue to lvalue seems to indicate that this would be viable.

theryee:
Still working on this.. noticed some interesting characteristics while writing some more test cases... hoping to get something posted this week.

rparolin:
no worries.  thanks for taking the time to do it right. :)

theryee:
I'm going to take a closer look later today... but I think I tripped over the MSVC 2013 is_copy_constructible is_move_constructible bug again...

theryee:
Still working on this. I've got a "slow" branch going that compiles in MSVC2013 but clang and gcc builds are an order of magnitude slower... not really sure why... I must've made a mistake somewhere.

theryee:
Ok-- sorted it all out.

include/EASTL/sort.h
```diff
@@ -213,7 +213,10 @@ namespace eastl
 	/// inner for loop with a call to upper_bound, which would be faster.
 	///
 	template <typename BidirectionalIterator, typename StrictWeakOrdering>
-	void insertion_sort(BidirectionalIterator first, BidirectionalIterator last, StrictWeakOrdering compare)
+	void insertion_sort(
```
rparolin: Please put all the parameters back on a single line

test/source/TestAlgorithm.cpp
```diff
@@ -116,6 +118,57 @@ struct TestObjectNegate : public eastl::unary_function<TestObject, TestObject>
 		{ return TestObject(-a.mX); }
 };
 
+struct MissingMoveConstructor
```
rparolin: Look at "ConceptImpls.h" these testing types are already added there and if one is missing it should be added there so its can be accessed via all tests.

test/source/TestAlgorithm.cpp
```diff
@@ -2339,22 +2392,83 @@ int TestAlgorithm()
 		}
 	}
 
+	// issue #92
```
rparolin: remove the github issue number comment

test/source/TestAlgorithm.cpp
```diff
@@ -2339,22 +2392,83 @@ int TestAlgorithm()
...
+					vec2.emplace_back(new int(randomNumber));
+					vec3.push_back(randomNumber);
+					vec4.push_back(randomNumber);
+				}
 
+				// used STL sort as reference for correctness
+				eastl::sort(vec1.begin(), vec1.end(), [](const eastl::unique_ptr<int>& lhs, const eastl::unique_ptr<int>& rhs) { return *lhs < *rhs; });
+				std::sort(vec2.begin(), vec2.end(), [](const eastl::unique_ptr<int>& lhs, const eastl::unique_ptr<int>& rhs) { return *lhs < *rhs; });
 
+				for(int i=0; i<numbersToSort; ++i){
```
rparolin: brackets formating

test/source/TestAlgorithm.cpp
```diff
@@ -2339,22 +2392,83 @@ int TestAlgorithm()
...
+					EATEST_VERIFY(*vec1[i] == *vec2[i]);
+				}
 
+				eastl::sort(vec3.begin(), vec3.end());
+				std::sort(vec4.begin(), vec4.end());
 
+				for(int i=0; i<numbersToSort; ++i){
```
rparolin: formatting

include/EASTL/sort.h
theryee: ok

test/source/TestAlgorithm.cpp
theryee: Ok. Moved them over. The structs I added are more like "negative" examples than the "positive" examples in that file.

test/source/TestAlgorithm.cpp
theryee: ok

test/source/TestAlgorithm.cpp
theryee: ok

test/source/TestAlgorithm.cpp
theryee: ok

theryee:
looks ok? think i got everything

include/EASTL/sort.h
```diff
@@ -868,14 +923,56 @@ namespace eastl
...
+
+		template <typename RandomAccessIterator, typename Size, typename Compare>
+		inline void quick_sort_impl(RandomAccessIterator first, RandomAccessIterator last, Size kRecursionCount, Compare compare,
+			typename eastl::enable_if<eastl::is_copy_constructible<typename iterator_traits<RandomAccessIterator>::value_type>::value>::type* = 0)
+		{
+			typedef typename iterator_traits<RandomAccessIterator>::value_type value_type;
+
+			while(((last - first) > kQuickSortLimit) && (kRecursionCount > 0))
+			{
+				const RandomAccessIterator position(eastl::get_partition<RandomAccessIterator, value_type, Compare>(first, last,
```
rparolin: Do we need to provide a copy-paste the implementations just to wrap in with eastl::forward?  Can't the default implementation use eastl::forward all the time?

include/EASTL/heap.h
```diff
@@ -144,6 +201,43 @@ namespace eastl
...
+	/// that is currently not within the heap.
+	///
+	template <typename RandomAccessIterator, typename Distance, typename T>
+	void adjust_heap(RandomAccessIterator first, Distance topPosition, Distance heapSize, Distance position, T&& value)
+	{
+		typedef typename iterator_traits<RandomAccessIterator>::value_type value_type;
+
+		// We do the conventional approach of moving the position down to the 
+		// bottom then inserting the value at the back and moving it up.
+		Distance childPosition = (2 * position) + 2;
```
rparolin: Same here.  Can't we have a single implementation?

include/EASTL/sort.h
theryee: I'll have to experiment with that this weekend. I seem to recall trying eastl::forward... but honestly it's been long enough that I don't remember if it was this particular function or a different one.

include/EASTL/heap.h
theryee: Ditto. Need to experiment with this again.

include/EASTL/sort.h
theryee: Ah I remember now. It's not the forward of the rvalue/lvalue that's the issue but the fact that one of them is using this: eastl::forward<const value_type> and the other is using east::forward<value_type>. The const variant can't always be used because the move constructor takes a non-const arg.

theryee:
Reduced code duplication. Lemme know what you think.

theryee:
I went ahead and employed same strategy for the code duplication removal... except for get_partition... in this case the code is not an exact duplicate...

theryee:
Ok I lied. Even get_partition can use the same strategy.

I think in theory even the comparator/non-comparator code duplication could be removed by adding a "comparator wrapper" class, e.g.
```
template<typename A, typename B>
class Comparator{
bool operator< (A, B){ return A < B; }
};
```

theryee:
Hey @rparolin, how does this look? Any more feedback? I'd like to close this PR out this week as my free time after this week is going to be limited.

include/EASTL/algorithm.h
```diff
@@ -939,6 +939,24 @@ namespace eastl
 		return eastl::make_pair(*iteratorPair.first, *iteratorPair.second);
 	}
 
+	template <typename T>
+	inline T median_impl(T a, T b, T c)
```
rparolin: They should be T&& (universal reference) to avoid copy-by-value?

include/EASTL/algorithm.h
```diff
@@ -976,23 +980,59 @@ namespace eastl
...
+	template <typename T>
+	inline T&& median(T&& a, T&& b, T&& c)
+	{
+		return eastl::forward<T>(median_impl<T&&>(eastl::forward<T>(a), eastl::forward<T>(b), eastl::forward<T>(c)));
+	}
+
+
 	template <typename T, typename Compare>
-	inline const T& median(const T& a, const T& b, const T& c, Compare compare)
+	inline T median_impl(T a, T b, T c, Compare compare)
```
rparolin: Same.  Take by T&&?

include/EASTL/heap.h
```diff
@@ -128,30 +172,58 @@ namespace eastl
...
+	/// new values into that vacated position appropriately. The value
+	/// argument is an entry which will be inserted into the heap after
+	/// we move nodes into the positions that were vacated.
 	///
 	/// This function requires that the value argument refer to a value
 	/// that is currently not within the heap.
 	///
-	template <typename RandomAccessIterator, typename Distance, typename T, typename Compare>
-	void adjust_heap(RandomAccessIterator first, Distance topPosition, Distance heapSize, Distance position, const T& value, Compare compare)
+	/// TODO: reduce code duplication here
```
rparolin: Is this todo still accurate?

include/EASTL/sort.h
```diff
@@ -213,7 +213,7 @@ namespace eastl
 	/// inner for loop with a call to upper_bound, which would be faster.
 	///
 	template <typename BidirectionalIterator, typename StrictWeakOrdering>
-	void insertion_sort(BidirectionalIterator first, BidirectionalIterator last, StrictWeakOrdering compare)
+	void insertion_sort( BidirectionalIterator first, BidirectionalIterator last, StrictWeakOrdering compare)
```
rparolin: whitespace change.  please revert.

include/EASTL/sort.h
```diff
@@ -223,7 +223,7 @@ namespace eastl
 
 			for(++iSorted; iSorted != last; ++iSorted)
 			{
-				const value_type temp(*iSorted);
+				value_type temp(eastl::forward<value_type>(*iSorted));
```
rparolin: why was const removed?

include/EASTL/sort.h
```diff
@@ -256,7 +255,7 @@ namespace eastl
...
```
rparolin: Why was const removed?

include/EASTL/sort.h
```diff
@@ -644,30 +643,21 @@ namespace eastl
...
-
-	/// get_partition
-	///
-	/// This function takes const T& instead of T because T may have special alignment
-	/// requirements and some compilers (e.g. VC++) are don't respect alignment requirements
-	/// for function arguments.
-	///
 	template <typename RandomAccessIterator, typename T>
-	inline RandomAccessIterator get_partition(RandomAccessIterator first, RandomAccessIterator last, const T& pivotValue)
+	inline RandomAccessIterator get_partition_impl(RandomAccessIterator first, RandomAccessIterator last, T pivotValue)
```
rparolin: Why is pivotValue no longer a reference?

include/EASTL/sort.h
```diff
@@ -678,24 +668,41 @@ namespace eastl
...
+
+	template <typename RandomAccessIterator, typename T>
+	inline RandomAccessIterator get_partition(RandomAccessIterator first, RandomAccessIterator last, T&& pivotValue)
+	{
+		// Note: unlike the copy-constructible variant of get_partition... we can't create a temporary const move-constructible object
+		return get_partition_impl<RandomAccessIterator, T&&>(first, last, eastl::move(pivotValue));
+	}
 
+	template <typename RandomAccessIterator, typename T, typename Compare>
+	inline RandomAccessIterator get_partition_impl(RandomAccessIterator first, RandomAccessIterator last, T pivotValue, Compare compare)
```
rparolin: pivotValue is a not a reference

include/EASTL/sort.h
```diff
@@ -797,10 +818,10 @@ namespace eastl
 			if(compare(*i, *first))
 			{
 				EASTL_VALIDATE_COMPARE(!compare(*first, *i)); // Validate that the compare function is sane.
-				const value_type temp(*i);
-				*i = *first;
+				value_type temp(eastl::forward<value_type>(*i));
```
rparolin: const was removed.

include/EASTL/sort.h
```diff
@@ -773,10 +794,10 @@ namespace eastl
 			if(*i < *first)
 			{
 				EASTL_VALIDATE_COMPARE(!(*first < *i)); // Validate that the compare function is sane.
-				const value_type temp(*i);
-				*i = *first;
+				value_type temp(eastl::forward<value_type>(*i));
```
rparolin: const removed.

include/EASTL/unique_ptr.h
```diff
@@ -336,9 +336,9 @@ namespace eastl
 		eastl::compressed_pair<pointer, deleter_type> mPair;
 
 		/// These functions are private in order to prevent copying, for safety.
-		unique_ptr(const this_type&);
-		unique_ptr& operator=(const this_type&);
-		unique_ptr& operator=(pointer pValue);
+		unique_ptr(const this_type&) = delete;
```
rparolin: Could you include this in a separate PR?  I don't feel like its related to the larger change.

include/EASTL/algorithm.h
theryee: Yes-- my intent was for these to not support copy-by-value. Will change.

include/EASTL/algorithm.h
theryee: will change.

include/EASTL/heap.h
theryee: no, will remove.

include/EASTL/sort.h
theryee: ok.

theryee:
Oh hah-- I didn't notice that the main CI failed but fork's CI didn't fail... I'll need to merge from master I guess.

include/EASTL/sort.h
theryee: The problem isn't really obvious unless you look down here on line 238:
```
*iNext = eastl::forward<value_type>(temp);
```

If `temp` is a `const value_type` and `iNext` is only move-assignable then assigning a const value is incompatible.

include/EASTL/sort.h
theryee: as above

include/EASTL/sort.h
theryee: will make this a universal reference to make clear what's occurring here.

include/EASTL/sort.h
theryee: as above.

include/EASTL/sort.h
theryee: Uh... don't remember why at this point... and I don't see why I did... added it back and everything compiles and benchmarks seem ok. Will add back.

theryee: This one doesn't fail unit tests... but that's an indicator that it needs a unit test... see below.

include/EASTL/sort.h
theryee: This is the same issue as before:
if `temp` is of type `const value_type` but `value_type` is only move-assignable, then when `adjust_heap` is invoked then the `temp` cannot be assigned.

include/EASTL/unique_ptr.h
theryee: will do

include/EASTL/algorithm.h
```diff
@@ -976,23 +980,59 @@ namespace eastl
 	/// Complexity: Either two or three comparisons will be required, depending 
 	/// on the values.
 	///
+	template <typename T>
+	inline T&& median(T&& a, T&& b, T&& c)
+	{
+		return eastl::forward<T>(median_impl<T&&>(eastl::forward<T>(a), eastl::forward<T>(b), eastl::forward<T>(c)));
```
rparolin: you shouldn't need the explict template parameter for median_impl here as it will be deduced from the forwarded function parameters

theryee: yup-- will remove. My intent was clarity at the cost of explicitness.

include/EASTL/sort.h
theryee: Still working on the unit test for this case... for some reason I can trigger the non-compare variant compile failure only via benchmarks but not unit tests.... investigating root cause...

Once I figure out why benchmarks and unit-tests invoke different code paths I'll add a new unit tests.

theryee:
Didn't merge and try buck yet... so this build might still fail...

include/EASTL/sort.h
theryee: Figured it out

theryee:
Not sure why buck build on linux is failing... is the job timing out? I don't see any obvious error messages:

https://travis-ci.org/theryee/EASTL/jobs/254678921#L1868

```
e/travis/.travis/job_stages: line 54: 79299 Killed                  buck run :test-runner
The command "if [[ -n "$USE_BUCK" ]]; then buck run :test-runner                                    ;fi" exited with 137.
0.00s$ if [[ ! -n "$USE_BUCK" ]]; then cd $TRAVIS_BUILD_DIR/build_$EASTL_CONFIG/test          ;fi
```

rparolin:
We can ignore that buck job.  I disabled it from the main CI since it was so flaky.

include/EASTL/sort.h
rparolin: Ah yes, we can't move from it if its const.

rparolin:
Thanks for the pull request!  :)

PR #102, opened 2017-04-30
Corrected the constructor of ref_count_sp

gian21391:
Corrected the definition of the constructor ref_count_sp to match the declaration


rparolin:
Nice catch.  Thanks.

Issue #103, opened 2017-05-16
Some of the member in red_black_tree could be declared as protected?


HenryRLee:
Some of the members in the red_black_tree file could be declared as protected, in my opinion.

These members include:

1. [mpNode](https://github.com/electronicarts/EASTL/blob/master/include/EASTL/internal/red_black_tree.h#L171)  in class rbtree_iterator
2. [the explicit constructor](https://github.com/electronicarts/EASTL/blob/master/include/EASTL/internal/red_black_tree.h#L175) of class rbtree_iterator
3. [mCompare](https://github.com/electronicarts/EASTL/blob/master/include/EASTL/internal/red_black_tree.h#L220) in class rb_base
4. [mAnchor, mnSize and mAllocator](https://github.com/electronicarts/EASTL/blob/master/include/EASTL/internal/red_black_tree.h#L367) in class rbtree

There is no failing code of this issue, but just some undesirable issues, that users can access variables that they shouldn't.
```
    eastl::set<int> myset;
    eastl::set<int>::iterator it;

    auto a0 = it.mpNode;
    auto a1 = myset.mCompare;
    auto a2 = myset.mAnchor;

```

rparolin:
the internal members you've highlighted I'd agree should have been protected.

i'm not sure the reasoning for the explicit constructor.

HenryRLee:
I think because the explicit constructor was using a type, which shouldn't be known by the user, as a parameter, so we shall keep the function protected.

PR #104, opened 2017-05-16
#103 Change some of the members in red_black_tree from public to protected

HenryRLee:
This commit is a solution to issue #103 
These changes include:
1. Change some members in red_black_tree from public to protected
2. Add friend declaration in class rbtree_iterator so that rbtree can access its private member
3. Some of the lines in map and set were using the iterator incorrectly. Instead of accessing mpNode and mValue, use the -> and * operators in the iterator directly.

### DIFF
--- a/include/EASTL/internal/red_black_tree.h
+++ b/include/EASTL/internal/red_black_tree.h
@@ -216,7 +216,7 @@ namespace eastl
 	{
 		typedef ExtractKey extract_key;
 
-	public:
+	protected:
 		Compare mCompare; // To do: Make sure that empty Compare classes go away via empty base optimizations.
 
 	public:
@@ -235,7 +235,7 @@ namespace eastl
 	{
 		typedef ExtractKey extract_key;
 
-	public:
+	protected:
 		Compare mCompare; // To do: Make sure that empty Compare classes go away via empty base optimizations.
 
 	public:
@@ -252,7 +252,7 @@ namespace eastl
 	{
 		typedef eastl::use_first<Pair> extract_key;
 
-	public:
+	protected:
 		Compare mCompare; // To do: Make sure that empty Compare classes go away via empty base optimizations.
 
 	public:
@@ -269,7 +269,7 @@ namespace eastl
 	{
 		typedef eastl::use_first<Pair> extract_key;
 
-	public:
+	protected:
 		Compare mCompare; // To do: Make sure that empty Compare classes go away via empty base optimizations.
 
 	public:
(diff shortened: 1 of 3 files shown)


include/EASTL/internal/red_black_tree.h
```diff
@@ -184,6 +184,22 @@ namespace eastl
 		rbtree_iterator& operator--();
 		rbtree_iterator  operator--(int);
 
+	protected:
+		template <typename Tp, typename PointerA, typename ReferenceA, typename PointerB, typename ReferenceB>
+		friend bool operator==(const rbtree_iterator<Tp, PointerA, ReferenceA>& a,
```
rparolin: Where are the implementations for these operators?

rparolin: Also, I'd rather use the class member function variant vs friending a freestanding functions.

rparolin: Also, why do you need to add these comparison operators?  Changing access control doesn't appear to warrant it.

rparolin:
Other than the comments about the friend function.   I think the changes look good on their own.  I'm not sure how much client code I will break by limiting access to these variables.

include/EASTL/internal/red_black_tree.h
HenryRLee: Hi, because these global functions already existed, between line 677 and 701. I just tried to make minimum changes to the original code.

HenryRLee: So basically they are friend forward declarations of these functions. https://github.com/electronicarts/EASTL/blob/master/include/EASTL/internal/red_black_tree.h#L675

rparolin: Why did they need to be added here?

HenryRLee: You mean why the original author wrote these functions? Because they wanted to have comparison between `iterator` and `const_iterator`.

rparolin: No, I mean why did you add the friend declarations for this change?

HenryRLee: Because these functions are accessing the internal node pointer, which is declared as protected now.

rparolin:
Apologises for not making this decision sooner (as I've gone back and forth on it) but I'm going to keep the code as-is for simplicity and legacy considerations.  Thanks for the work and apologises again.

HenryRLee:
I see. No problem.

PR #106, opened 2017-05-24
Add support for Android on ARM64

mwolting:
The library previously did not recognize Android on ARM64 targets, and did not support breakpoints even if the target were to be recognized. Fixed by adding the required definitions in EABase/config/eaplatform.h, and adding a correct breakpoint instruction to both EASTL and EATest.


rparolin:
Awesome.  Thanks.

PR #110, opened 2017-06-12
Buck build support

njlr:
This PR adds support for builds using [Buck](https://buckbuild.com/). Buck is a build tool by Facebook that allows for faster builds, reproducibility and easier composition of libraries. 

To build EASTL using Buck:

```bash=
buck build :eastl
```

To run the tests:

```bash=
buck run :test-runner
```

To run the benchmarks:

```bash=
buck run :benchmark
```

The existing CMake build is unchanged; the two can coexist peacefully 😊


rparolin:
Cool.  Could you add a travis ci step so that we don't regress into broken builds when  using Buck.

njlr:
Sure, I'll have a go.

njlr:
It took a few tries, but Travis will now run the Buck build.

.travis.yml
```diff
@@ -60,9 +88,21 @@ before_script:
...
 
 script:
-    # Run Tests 
-    - cd $TRAVIS_BUILD_DIR/build_$EASTL_CONFIG/test
-    - ctest -C $EASTL_CONFIG -V || exit 1
-    # Run Benchmarks
-    - cd $TRAVIS_BUILD_DIR/build_$EASTL_CONFIG/benchmark
-    - ctest -C $EASTL_CONFIG -V || exit 1
+  # Run the Buck build and tests
+  - buck build :eastl
```
rparolin: rather than doing the buck builds inline with the cmake builds... lets make a seperate travis task for them... similar to to the above for clang3.9

njlr:
The Buck builds are now run as separate jobs on Travis CI.

.travis.yml
```diff
@@ -4,65 +4,113 @@ os:
...
     include:
         - compiler: clang "debug build with clang3.9"
           env: EASTL_CONFIG=Debug   USE_CLANG_39=yes
           os: osx
-        - compiler: clang  "release build with clang3.9"
+        - compiler: clang "release build with clang3.9"
           env: EASTL_CONFIG=Release USE_CLANG_39=yes
           os: osx
+        - compiler: clang "build with Buck"
+          env: USE_BUCK=yes
```
rparolin: debug or release builds?  Should we have both?  Or just default to one or the other?

rparolin:
The GCC build was 17+mins.   Is this normal or to be expected?  Or will all future builds being shorter because the objs are cached?

njlr:
Buck requires GCC 5 or greater, so that extra time is installing a later GCC using Linuxbrew. I didn't want to change the GCC version for apt-get because that would affect the other builds, so this seemed like the best way. The Linuxbrew folder is cached so the installation is skipped for subsequent builds.

The Buck build itself is very quick (just over 2 minutes on OSX, where we use Clang not GCC).

If it's too long, perhaps we could only run Buck on OSX? Or maybe we could move all builds to a later GCC? 4.8 is quite old.

On 20 Jun 2017, at 04:13, Rob Parolin <notifications@github.com<mailto:notifications@github.com>> wrote:


The GCC build was 17+mins. Is this normal or to be expected? Or will all future builds being shorter because the objs are cached?

.travis.yml
njlr: It currently only does a release build. Buck does not have built-in release and debug targets like CMake does, however they are [easy to implement](https://github.com/njlr/buck-custom-flavours/blob/master/.buckconfig). 

Is this a requirement? Happy to add it to the PR.

rparolin: It can be added later if required.  Many thanks for the PR.

rparolin:
https://travis-ci.org/electronicarts/EASTL/jobs/245451895

The GCC install appears very fast actually.   There seems to be an intermintent issue where the GCC builds of the test cpp will hang and force a timeout.  Could you look at this?

njlr:
This is difficult for me to debug because I am macOS. 

Since there is no difference in the `BUCK` files for Linux+GCC vs OSX+Clang, perhaps the best solution would be to remove the Buck build on Travix+Linux untill there is demand? 

[Here is the code for this](https://github.com/njlr/EASTL/commit/ae7f19979259ce0b69c2f69905ee4162a07922cc). Happy to send a PR.

swardle:
Hey Nick

Another option would be to use Oracle virtual box on mac.  It also works on pc.   It is a bit slow on my mac 11" air mind you but probably fast enough to fix some small build problems on Linux gcc. 

Scott

rparolin:
I tried to get buck running on Win10  via Linux Subsystem but couldn't get
buck stood up due to JDK versiob issues.  VM is probably best solution.

Issue #115, opened 2017-07-18
Warnings related to missuse of pragmas and litteral operator suffixes not preceded by '_'


kfitch:
Using EASTL 3.05.07
When compiling with gcc v5.4.0 on linux I get a warnings like

    include/EASTL/internal/config.h:1797:24: warning: unknown option after '#pragma GCC diagnostic' kind [-Wpragmas]
      EA_DISABLE_GCC_WARNING(-Wno-literal-suffix)

As well as warnings like

    include/EASTL/string.h:3749:30: warning: literal operator suffixes not preceded by '_' are reserved for future standardization

I get similar warnings when using gcc 7.1.0.

BUT, if I change line 1797 of config.h to 

      EA_DISABLE_GCC_WARNING(-Wliteral-suffix)

then the pragma warnings on gcc5.4 go away, but unfortunately the "future standaradization" warning stays. On gcc7.1 both warnings go away when the "no-" is removed.

rparolin:
Thanks for submitting this.  Sounds like I just messed up and copy-pasted the clang suppression warning for gcc.  Would you like to submit a PR that fixes this issue?

PR #117, opened 2017-08-15
Fix pragma to disable "literal suffix" warning on GCC.

kfitch:
Unfortunately this only disables the warning on GCC 7.x. GCC 6.x seems to be
unable to disable this warning.

See Issue #115


rparolin:
Thanks!

PR #121, opened 2017-09-27
define _SSIZE_T_DEFINED, when defining typedef ssize_t

Leandros:
Not defining that ssize_t has already been typedef'ed might upset other libraries trying to define it, too. The `_SSIZE_T_DEFINED` macro is taken from ZeroMQ, since their checking for it. Considering ZeroMQ is a pretty widespread & successfully deployed library, I'd consider using that name safe.
Without it, ZeroMQ fails to compile, due to conflicting types used for ssize_t.

### DIFF
--- a/test/packages/EABase/include/Common/EABase/eabase.h
+++ b/test/packages/EABase/include/Common/EABase/eabase.h
@@ -587,12 +587,15 @@
 	// As of this writing, all non-GCC compilers significant to us implement 
 	// uintptr_t the same as size_t. However, this isn't guaranteed to be 
 	// so for all compilers, as size_t may be based on int, long, or long long.
-	#if defined(_MSC_VER) && (EA_PLATFORM_PTR_SIZE == 8)
-        #define _SSIZE_T_DEFINED
-		typedef __int64 ssize_t;
-	#else
-        #define _SSIZE_T_DEFINED
-		typedef long ssize_t;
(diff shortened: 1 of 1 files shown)


rparolin:
Can you provide supporting documentation for the '_SSIZE_T_DEFINED' preprocessor macro?

Leandros:
I can't provide any documentation, but I found that `_SSIZE_T_DEFINED` is the de-facto guarding macro used for defining `ssize_t` in open source projects. Examples of usage:

- [czmq](https://github.com/zeromq/czmq/blob/master/include/czmq_prelude.h#L542)
- [node.js](https://github.com/nodejs/node/blob/master/src/node.h#L185)
- [libuv](https://github.com/libuv/libuv/blob/2e5ad25cadd9c12463f4150f04a6e99f40c3f880/include/uv-win.h#L29)
- [mingw-64 on win32](https://github.com/mirror/mingw-w64/blob/master/mingw-w64-headers/crt/_cygwin.h#L20)

to just name a few. 

To be even more safe, one could also add `_SSIZE_T_` (as seen by node.js & libuv) & `__ssize_t_defined` (us used by the glibc / gcc headers).

rparolin:
re-running ci validation on pr

rparolin:
Thanks for the PR! :)

Issue #113, opened 2017-07-15
eastl::remove() and remove_if() use copy assignment


Vuhdo:
Hi! While trying to adapt our codebase to use this great library, I came across one specific issue: `eastl::remove_if()` uses copy assignment operator when swapping values. Looks like a little bit ineffective. I checked two major STL implementations and they both use `std::move()` inside `std::remove_if()`.

rparolin:
thanks for reporting

rparolin:
The fix for this should be in the next EASTL release.

PR #122, opened 2017-10-05
+ bugfix with remove_copy_if when using remove_if with vector<unique_ptr<>>

djdron:


djdron:
seems to be duplicate of #113

PR #123, opened 2017-10-08
Rename 'IN' to 'in', to avoid compiler errors on windows

Leandros:
`IN` is defined to nothing on Win32, it used to be a predecessor to SAL, as described here: https://docs.microsoft.com/en-us/visualstudio/code-quality/best-practices-and-examples-sal

Using `in` instead, allows compilation of EASTL with `clang` on Win32.


rparolin:
Unfortunately, I can't merge the PR because of the broken CI.  I'll have to look at it.

Leandros:
The CI issue looks like that `clang++` is trying to use GCCs libstdc++, as it is pointing towards these headers. Possibly fixable by adding `-stdlib=libc++`.

rparolin:
closing an reopening to rerun PR validation.

PR #124, opened 2017-10-12
Check for empty range in vector::erase(first,last)

bnason-nf:
When erase() is called on a vector with an empty range (first == last),
it should be a no-op. If erase() does not check for this case, it can
end up calling the move assignment operator for an obect with itself as
the source and dest. This is unnecessary work at best, and at worst
could cause problems if the object's implementation of the move
assignment does not handle this case well, leaving the object in an
undefined state.

This fix simply adds a check for first==last and avoids doing any work
in that case.

### DIFF
--- a/include/EASTL/vector.h
+++ b/include/EASTL/vector.h
@@ -1283,9 +1283,12 @@ namespace eastl
 				EASTL_FAIL_MSG("vector::erase -- invalid position");
 		#endif
  
-		iterator const position = const_cast<value_type*>(eastl::move(const_cast<value_type*>(last), const_cast<value_type*>(mpEnd), const_cast<value_type*>(first)));
-		eastl::destruct(position, mpEnd);
-		mpEnd -= (last - first);
+		if (first != last)
+		{
+			iterator const position = const_cast<value_type*>(eastl::move(const_cast<value_type*>(last), const_cast<value_type*>(mpEnd), const_cast<value_type*>(first)));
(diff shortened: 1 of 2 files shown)


test/source/TestVector.cpp
```diff
@@ -144,6 +144,15 @@ struct testmovable
 	testmovable& operator=(testmovable&&) EA_NOEXCEPT { return *this; }
 };
 
+#if EASTL_MOVE_SEMANTICS_ENABLED
+struct testmovetoself
```
rparolin: Please camel case the struct name.  The class above shouldn't be like that. I'll fix that.

rparolin: Not crazy about the name b/c its actually not testing if its moved to self.  Its really just setting a flag if the copy-assignment operator is called.  I'd either have this class test for self assignment or rename it so the intent is clear.

include/EASTL/vector.h
```diff
@@ -1283,9 +1283,12 @@ namespace eastl
 				EASTL_FAIL_MSG("vector::erase -- invalid position");
 		#endif
  
-		iterator const position = const_cast<value_type*>(eastl::move(const_cast<value_type*>(last), const_cast<value_type*>(mpEnd), const_cast<value_type*>(first)));
-		eastl::destruct(position, mpEnd);
-		mpEnd -= (last - first);
+		if (first != last)
```
rparolin: Good catch!

PR #129, opened 2017-11-10
 just some even prettier alignment and a tiny typo in comments fixed

sfinktah:
I don't even know if you accept PR's, you certainly don't need to accept this one :)

Just saw how nice your alignment was, and saw a block that looked sad...  and a silly typo

### DIFF
--- a/include/EASTL/bonus/ring_buffer.h
+++ b/include/EASTL/bonus/ring_buffer.h
@@ -229,7 +229,7 @@ namespace eastl
 		size_type          mSize;
 
 	public:
-		// There currently isn't a ring_buffer constructor that specifies and initial size, unlike other containers.
+		// There currently isn't a ring_buffer constructor that specifies an initial size, unlike other containers.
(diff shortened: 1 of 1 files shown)


sfinktah:
Least I can do, considering how up-to-date this repo is kept.  Do you mind if I ask a question?

What code editor(s) do you use?

I ask, because I use VS2017, and it's formatting (and re-formatting) is god-awful and can't be disabled for sections.  I switch to `vim` when I actually need to align something manually, and try not to mess it up otherwise.   If there was an alternative that had a vim-emulation plugin, and the equiv. of intellisense (that worked), I'd be interested.

theryee:
I use visual studio code. It has a vim emulation plug-in, decent intellisense like feature, and a clang format plugin.

rparolin:
I'm a vim user and there is a clang format file to help with formatting
issues.  Its not perfect though... but good enough I'd recommend its use.

PR #131, opened 2017-11-29
Make return types of insert conforming with c++11

wmamrak:
Make return types of insert conforming with c++11, i.e. iterator, instead of void.


rparolin:
Looks good.  Can you add a unit test for this?

wmamrak:
Sure

PR #133, opened 2017-12-04
 Make return types of insert conforming with c++11 (with tests)

wmamrak:


rparolin:
Thanks for the PR. :)

Issue #119, opened 2017-09-11
deque::shrink_to_fit clears all data


Buracula:
```
template <typename T, typename Allocator, unsigned kDequeSubarraySize>
inline void deque<T, Allocator, kDequeSubarraySize>::shrink_to_fit()
{		
	return set_capacity(0);
}
```
If we use shrink_to_fit on deque it clears all items instead of deallocating unnecessary memory.

rparolin:
Sorry for the delay.  Thanks for reporting this issue.

theryee:
Does shrink_to_fit need to do anything for the EASTL implementation?

pop_back seems to free the subarray's when no longer required:
```
			// This is executed only when we are popping the first item off the last subarray.
			// In this case we need to free the subarray and point mItEnd to the previous subarray.
```

rparolin:
I implemented this and it will be in the next EASTL release.

PR #134, opened 2017-12-20
shrink_to_fit issue#119

theryee:
issue #119


rparolin:
Thanks for the PR.   Unfortunate that did the wrong thing.

PR #137, opened 2018-01-03
minor faq issue fix

claimred:
Though I feel like there should be some other argument in favor of 

> The STL has specifications that limit our ability to use it efficiently


rparolin:
Thanks!

PR #142, opened 2018-02-18
String view hash fixes

artemartey:
fixes for string_view hash calc implementation

### DIFF
--- a/include/EASTL/string_view.h
+++ b/include/EASTL/string_view.h
@@ -458,11 +458,11 @@ namespace eastl
 	{
 		size_t operator()(const string_view& x) const
 		{
-			const unsigned char* p = (const unsigned char*)x.data(); // To consider: limit p to at most 256 chars.
-			unsigned int c, result = 2166136261U; // We implement an FNV-like string hash. 
-			while((c = *p++) != 0) // Using '!=' disables compiler warnings.
-				result = (result * 16777619) ^ c;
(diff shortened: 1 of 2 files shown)


include/EASTL/string_view.h
```diff
@@ -458,35 +458,35 @@ namespace eastl
 	{
 		size_t operator()(const string_view& x) const
 		{
-			const unsigned char* p = (const unsigned char*)x.data(); // To consider: limit p to at most 256 chars.
-			unsigned int c, result = 2166136261U; // We implement an FNV-like string hash. 
-			while((c = *p++) != 0) // Using '!=' disables compiler warnings.
-				result = (result * 16777619) ^ c;
-			return (size_t)result;
+			string_view::const_iterator p = x.cbegin();
+			size_t result = 2166136261U; // We implement an FNV-like string hash.
```
virt00l: Unfortunately this implicitly changed intermediate type of 'result' variable  from 32-bit of 'unsigned' type to 64-bit size_t (on 64-bit platforms, obviously). But hash<const char*> still calculates in unsigned's! Which not only causing different 
results for same "const char*"  & string_view that was made from that "const char*", but different hashes for the same strings across 32/64 platforms. 

We had to revert this patch internally.

artemartey: I fix it here #143.  Please review.

PR #143, opened 2018-02-19
String view type fixes

artemartey:


include/EASTL/string_view.h
```diff
@@ -459,10 +459,10 @@ namespace eastl
 		size_t operator()(const string_view& x) const
 		{
 			string_view::const_iterator p = x.cbegin();
-			size_t result = 2166136261U; // We implement an FNV-like string hash.
+			uint32_t result = 2166136261U; // We implement an FNV-like string hash.
```
virt00l: This is looks like it will work, altough I not verified it yet.

include/EASTL/string_view.h
```diff
@@ -459,10 +459,10 @@ namespace eastl
...
 			while (p != x.cend())
```
virt00l: I would consider to extract (i.e. to cache in local var) 'x.cend()' from loop (as it's actually loop invariant, i.e. is not changing within loop). Some compilers are probably smart enough to do it automatically, for others - it is need to be done manually

artemartey:
all compilers I knew can make these optimizations, but maybe you're right.

rparolin:
Looks good.  Thanks!

PR #144, opened 2018-02-19
fix eastl::optional<> MSVC /Wall level 4 warnings

virt00l:
optional.h(42): warning C4574: 'EASTL_EXCEPTIONS_ENABLED' is defined to be '0': did you mean to use '#if EASTL_EXCEPTIONS_ENABLED'?
optional.h(81): warning C4582: 'eastl::Internal::optional_storage<T,false>::val': constructor is not implicitly called


rparolin:

PR #145, opened 2018-02-19
fix broken include numeric.h dependency on iterator.h

virt00l:
If numeric.h included before iterator.h was included it won't compile
on GCC, e.g:

EASTL/numeric.h:132:35: error: expected initializer before ¿<¿ token typedef typename iterator_traits<InputIterator>::value_type value_type;

and Clang, e.g.:

EASTL/numeric.h:159:20: error: expected a qualified name after 'typename'
                typedef typename iterator_traits<InputIterator>::value_type value_type;

### DIFF
--- a/include/EASTL/numeric.h
+++ b/include/EASTL/numeric.h
@@ -13,6 +13,7 @@
 
 
 #include <EASTL/internal/config.h>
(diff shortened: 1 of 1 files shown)


rparolin:
Thanks!

PR #146, opened 2018-02-22
fix eastl::string_view compare for string with different sizes

zheland:
string_view.compare method should return non-zero result,
when compare strings with different sizes and same common part.

Added helper static compare function similar to used in string.h.
Fixed string_view compare method using static compare function.
Fixed and added unit tests for string_view.

### DIFF
--- a/include/EASTL/string_view.h
+++ b/include/EASTL/string_view.h
@@ -158,9 +158,19 @@ namespace eastl
 			return this_type(mpBegin + pos, count);
 		}
 
-		EA_CONSTEXPR int compare(basic_string_view sw) const EA_NOEXCEPT
+		static EA_CPP14_CONSTEXPR int compare(const T* pBegin1, const T* pEnd1, const T* pBegin2, const T* pEnd2)
 		{
-			return Compare(mpBegin, sw.data(), eastl::min_alt(size(), sw.size()));
+			const ptrdiff_t n1   = pEnd1 - pBegin1;
+			const ptrdiff_t n2   = pEnd2 - pBegin2;
+			const ptrdiff_t nMin = eastl::min_alt(n1, n2);
+			const int       cmp  = Compare(pBegin1, pBegin2, (size_t)nMin);
+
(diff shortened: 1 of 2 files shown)


virt00l:
@rparolin Please review. Without this PR (i.e. as it is right now in master) string_view<> sort is essentially broken (which means that broken any attempt to use it in various containers like set<>, map<>, etc...)

include/EASTL/string_view.h
```diff
@@ -158,9 +158,19 @@ namespace eastl
...
-		EA_CONSTEXPR int compare(basic_string_view sw) const EA_NOEXCEPT
+		static EA_CPP14_CONSTEXPR int compare(const T* pBegin1, const T* pEnd1, const T* pBegin2, const T* pEnd2)
 		{
-			return Compare(mpBegin, sw.data(), eastl::min_alt(size(), sw.size()));
+			const ptrdiff_t n1   = pEnd1 - pBegin1;
+			const ptrdiff_t n2   = pEnd2 - pBegin2;
+			const ptrdiff_t nMin = eastl::min_alt(n1, n2);
+			const int       cmp  = Compare(pBegin1, pBegin2, (size_t)nMin);
+
+			return (cmp != 0 ? cmp : (n1 < n2 ? -1 : (n1 > n2 ? 1 : 0)));
```
rparolin: The return value of compare should be taking care of this.  Why are these conditionals necessary?

virt00l: Because it *is* 'compare' of string_view. If yor meant 'Compare' instead than it's defined in char_traits.h and it's full implementation is
```
inline int Compare(const char8_t* p1, const char8_t* p2, size_t n)
	{
		return memcmp(p1, p2, n);
	}
```
which oviously works right only for strings of equal length. 
Also this code is what *precisely* eastl::string does as well, see
https://github.com/electronicarts/EASTL/blob/master/include/EASTL/string.h#L2970

rparolin: Ah gotcha.  Thanks.

PR #156, opened 2018-04-24
Bazel build support

mapx:
Bazel build support as https://github.com/electronicarts/EASTL/pull/110 did.

This may not be of the best Bazel style, but each `BUCK` file has a relative `BUILD` file so that maintenance cost would keep low.

`.travis.yml` file is not changed yet.

Tested environment:
Linux x86_64 with gcc-6.4.0 and gcc-7.3.0


mapx:
That's all right.

I'm using EASTL with Bazel build and will maintain the fork. It might interest you after it's battle-tested for a long time.

Love your great work!

mapx:
The link is https://github.com/mapx/EASTL

I'll follow your updates and patch Bazel files in the fock :)

PR #162, opened 2018-05-04
Fixing eastl::bitset natvis

### LINKED ISSUE #154: Bug in eastl::bitset natvis ?
Hello,

It looks like the natvis for eastl::bitset doesn't show the right values (or at least, I don't understand what it shows). 

For example:
```c++
eastl::bitset<4> a;
a.set(3);
```
shows `{ false, true, true, true }` instead of `{ false, false, false, true }`.

The reason for that, I think, is a missing mask [on this line](https://github.com/electronicarts/EASTL/blob/master/doc/EASTL.natvis#L428) to make it test only one bit, instead of all the remaining bits in the word. The bit 0 is correct however, because it's handled by a special case that uses a mask (not sure why).

Replacing the code by this, makes it display what I expect:
```xml
<Loop>
	<Exec>bBitValue = ((mWord[iWord] >> iBitInWord) % 2) != 0 ? true : false</Exec>
	<Item>bBitValue</Item>
	<Exec>iBitInWord++</Exec>
	<If Condition="iBitInWord == kBitsPerWord">
		<Exec>iWord++</Exec>
		<Exec>iBitInWord = 0</Exec>
	</If>
</Loop>
```

Is it a bug? And if yes, do you want a pull request with this change?

jlaumon:
Fixes #154, the bitset natvis wasn't showing correct values.

image

### DIFF
--- a/doc/EASTL.natvis
+++ b/doc/EASTL.natvis
@@ -421,12 +421,7 @@
       <Size>kSize</Size>
 
       <Loop>
-        <If Condition="iBitInWord == 0">
-          <Exec>bBitValue = (mWord[iWord] % 2 != 0) ? true : false</Exec>
-        </If>
-        <Else>
-          <Exec>bBitValue = (mWord[iWord] >> iBitInWord) != 0 ? true : false</Exec>
-        </Else>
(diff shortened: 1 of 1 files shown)


rparolin:
Awesome.  Thanks for the PR!

PR #165, opened 2018-05-17
Correct the name of container's reset member function

mapx:
I might be wrong, but it turns out the source code uses `reset_lose_memory()` instead of `reset()` in question?


rparolin:
Thanks for the PR.

PR #172, opened 2018-06-02
Radix sort benchmark improvements

dave-cope:
Expanded the sorting benchmarks.  The change adds a new RandomizationType and also adds support for radix_sort in the "slow assignment" benchmark.

Additionally this change simplifies some of the radix_sort code so that data can be sorted without first being copied to a temporary buffer and back.


rparolin:
Looks good.  Thanks.

PR #174, opened 2018-06-12
Radix sort test improvements

dave-cope:
This PR adds a test which demonstrates a bug in radix_sort, and also fixes the bug (in a separate commit).


rparolin:
Awesome.  Thanks for the PR.

PR #176, opened 2018-06-15
Fixed a bug related to gVerbosityLevel.

dave-cope:
The variable was previously declared as being static which meant it could take on different values in different translation units.  The issue was resolved by making the variable extern and defining it in EAEntryPointMain.inl (which is assumed to be included exactly once for a program).


rparolin:
Thanks!

PR #177, opened 2018-06-18
Fix constexpr specifier for integral_constant

zheland:
integral_constant<T, v>::value should be constexpr.

This fix use of constexpr non-integral constants, e.g.:
type_traits.h:264:18: error: `constexpr` needed
for in-class initialization of static data member
`int* const eastl::integral_constant<int*, 0u>::value`
of non-integral type [-fpermissive]

### DIFF
--- a/include/EASTL/type_traits.h
+++ b/include/EASTL/type_traits.h
@@ -261,7 +261,7 @@ namespace eastl
 	template <typename T, T v>
 	struct integral_constant
 	{
-		static const T value = v;
+		static EA_CONSTEXPR T value = v;
 		typedef T value_type;
 		typedef integral_constant<T, v> type;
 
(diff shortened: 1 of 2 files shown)


zheland:
From C++14 draft N4296:
```
template <class T, T v>
struct integral_constant {
static constexpr T value = v;
```

From VisualStudio 14.0 xtr1common header:
```
template<class _Ty,
	_Ty _Val>
	struct integral_constant
	{	// convenient template for integral constant types
	static constexpr _Ty value = _Val;
```

From GCC type_traits header:
```
  template<typename _Tp, _Tp __v>
    struct integral_constant
    {
      static constexpr _Tp                  value = __v;
```

rparolin:
Good change.  I'll merge once the CI goes green.

PR #178, opened 2018-06-18
Adding "inline" to get(...) functions for TupleImpl

CypherSignal:
The lack of inline on these functions was causing some performance problems in some situations where tuple is heavily utilized, often resulting in having to change Cmake's "RelWithDebugInfo" generated configuration in order to get accurate performance data.

e.g. a simple loop over a vector of tuples such as...
```
	int numTupleBools = 0;	
	for (auto& iter : tupleVec)
	{
		numTupleBools += get<0>(iter) ? 1 : 0;
	}
```
...resulted in nearly twice the number of instructions and execution time, due to having to call eastl::Internal::get(...) on each iteration, instead of accessing the data in the tuple directly.

Note that both of these functions callers and callees have the inline tag, and when "any suitable" inlining is enabled, these functions get inlined away in almost all cases - after all, it just acts as an intermediary between TupleImpl and one of its TupleLeaf's.

### DIFF
--- a/include/EASTL/tuple.h
+++ b/include/EASTL/tuple.h
@@ -396,42 +396,42 @@ struct TupleImpl<integer_sequence<size_t, Indices...>, Ts...> : public TupleLeaf
 };
 
 template <size_t I, typename Indices, typename... Ts>
-tuple_element_t<I, TupleImpl<Indices, Ts...>>& get(TupleImpl<Indices, Ts...>& t)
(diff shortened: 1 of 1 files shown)


rparolin:
Nice.  I'll merge when the CI goes green.  Not sure what is happening there.

rparolin:
I can manually re-kick the validation on the PR.   But I don't know if (I doubt it) if changes need to made to the yml file to fix this.

rparolin:
Thanks for the PR!

PR #179, opened 2018-06-20
Made a couple radix sort optimizations

dave-cope:
This PR adds some optimizations for radix_sort.  The description below comes from the comments in the modified radix_sort implementation.

----

The radix_sort implementation uses two optimizations that are not part of a typical radix sort implementation.
1. Computing a histogram (i.e. finding the number of elements per bucket) for the next pass is done in parallel with the loop that "scatters" elements in the current pass.  The advantage is that it avoids the memory traffic / cache pressure of reading keys in a separate operation. **Note**: It would also be possible to compute all histograms in a single pass.  However, that would increase the amount of stack space used and also increase cache pressure slightly.  However, it could still be faster under some situations.
2. If all elements are mapped to a single bucket, then there is no need to perform a scatter operation.  Instead the elements are left in place and only copied if they need to be copied to the final output buffer.

### DIFF
--- a/include/EASTL/sort.h
+++ b/include/EASTL/sort.h
@@ -1600,116 +1600,116 @@ namespace eastl
 				{ return x.mKey; }
 		};
 
-		// uint8_t version.
-		template <typename RandomAccessIterator, typename ExtractKey>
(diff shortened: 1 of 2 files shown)


rparolin:
Thanks Dave!

PR #185, opened 2018-06-23
Adding tests and fixes for, e.g., comparison of tuples w/ non-const refs

CypherSignal:
`const` qualifier on `const tuple_element_t<I, TupleImpl<Indices, Ts...>>&` gets dropped when Ts is an lvalue ref. In this case, the only source of 'const-ness' on `tuple_element_t` comes from being baked into the source type. This means for non-const lvalue refs, the return type just becomes `Ts&`, but `getInternal` will attempt to return a const-ref value, and fail to compile because we cannot remove the const qualifier.

This problem was discovered when trying to use a comparison operator on a tuple containing non-const refs: one of the functions specifically coerces the input parameters to be `const tuple<Ts...>&`, and attempted to then call `get<...>(const tuple<Ts...>&)` on it. This was avoided until a recent experiment because the object used for the input was not strictly const, or did not contain non-const ref types. 

Attempted alternatives:

- Doing a `const_cast` to coerce the return from `getInternal` to not be `const`, and then have it be switched back to `const` or not as needed (for lvalue-refs, or just value types) works. It works, but does not fix the problem of the return type simply being wrong.

- Setting `tupleImpl` to be `const`, i.e. `tuple_element_t<I, const TupleImpl<Indices, Ts...>`, does get the return type correct, but elsewhere the compiler attempts to instantiate the types for the function when evaluating overloads for `tuple`'s that include move-only datatypes, and compilation fails because of attempts to add const to an rvalue object. Doesn't work.

- Setting `Ts` itself to be `const` in the `tuple_element_t` declarations failed because the `const` was still getting stripped off at some point. Doesn't work.

- Other transformations to how `tuple_element_t` devolves into a type - even using `typename tuple_element<...>` directly just resulted in a similar scenario. Trying to manually tack on `const` to the type in question after the fact, when we're dealing with changing the ref to be const and not changing the _type of_ the ref to be const, simply didn't work.

For more information: https://en.cppreference.com/w/cpp/language/reference

> Reference types cannot be cv-qualified at the top level; there is no syntax for that in declaration, and if a qualification is introduced through a typedef, decltype, or template type argument, it is ignored.

### DIFF
--- a/include/EASTL/internal/tuple_fwd_decls.h
+++ b/include/EASTL/internal/tuple_fwd_decls.h
@@ -23,12 +23,20 @@ namespace eastl
 	template <size_t I, typename Tuple>
 	using tuple_element_t = typename tuple_element<I, Tuple>::type;
 
+	// const typename for tuple_element_t, for when tuple or TupleImpl cannot itself be const
+	template <size_t I, typename Tuple>
+	using const_tuple_element_t = typename conditional<
+						is_lvalue_reference<tuple_element_t<I, Tuple>>::value,
+							 add_lvalue_reference_t<const remove_reference_t<tuple_element_t<I, Tuple>>>,
+							 const tuple_element_t<I, Tuple>
+						>::type;
+
 	// get
 	template <size_t I, typename... Ts_>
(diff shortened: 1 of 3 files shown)


rparolin:
Awesome thanks!

PR #187, opened 2018-06-25
C++14: type_trait is_final

theryee:
Add C++14's is_final

### DIFF
--- a/include/EASTL/internal/type_compound.h
+++ b/include/EASTL/internal/type_compound.h
@@ -715,6 +715,10 @@ namespace eastl
 
 	#endif
 
+	// is_final requires compiler support detect final classes: __is_final
+	template <typename T>
+	struct is_final : public integral_constant<bool, __is_final(T)> {};
+
 } // namespace eastl
 
 
--- a/include/EASTL/internal/type_compound.h
+++ b/include/EASTL/internal/type_compound.h
@@ -715,9 +715,11 @@ namespace eastl
 
 	#endif
 
-	// is_final requires compiler support detect final classes: __is_final
-	template <typename T>
-	struct is_final : public integral_constant<bool, __is_final(T)> {};
+	#ifdef EASTL_COMPILER_INTRINSIC_TYPE_TRAITS_AVAILABLE
+		// is_final requires compiler support detect final classes: __is_final
+		template <typename T>
+		struct is_final : public integral_constant<bool, __is_final(T)> {};
+	#endif
 
 } // namespace eastl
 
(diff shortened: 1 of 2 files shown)


test/source/TestTypeTraits.cpp
```diff
@@ -1823,6 +1831,20 @@ int TestTypeTraits()
...
 
+	// is_final
+	static_assert(std::is_final<FinalStruct>::value, "final struct not correctly detected");
+	static_assert(std::is_final<FinalClass>::value, "final struct not correctly detected");
+	static_assert(!std::is_final<Enum>::value, "enum not correctly detected");
+	static_assert(!std::is_final<int>::value, "int not correctly detected");
+	static_assert(!std::is_final<Struct>::value, "non-final struct not correctly detected");
+	static_assert(!std::is_final<Class>::value, "non-final class not correctly detected");
+	static_assert(is_final<FinalStruct>::value, "final struct not correctly detected");
+	static_assert(is_final<FinalClass>::value, "final struct not correctly detected");
```
rparolin: We should test that our type_trait returns the same answer as the platforms instead of testing them seperately.

include/EASTL/internal/type_compound.h
```diff
@@ -710,6 +710,12 @@ namespace eastl
 
 	#endif
 
+	#ifdef EASTL_COMPILER_INTRINSIC_TYPE_TRAITS_AVAILABLE
+		// is_final requires compiler support detect final classes: __is_final
+		template <typename T>
+		struct is_final : public integral_constant<bool, __is_final(T)> {};
```
rparolin: We should provide a default implementation so that use code doesn't need to create seperate code paths.  If the compiler instrinic isn't available have is_final return false_type.

theryee: Sure, that makes sense.

test/source/TestTypeTraits.cpp
theryee: Sure, makes sense.

theryee:
Good to go?

rparolin:
Sorry.  I made comments but never submitted them.

test/source/TestTypeTraits.cpp
```diff
@@ -1823,6 +1831,15 @@ int TestTypeTraits()
 	}
 
 	static_assert(gIntNullptrConstant == nullptr, "integral constant with value of nullptr failure");
+	
+	// is_final
+
+	static_assert(std::is_final<FinalStruct>::value == is_final<FinalStruct>::value, "final struct not correctly detected");
```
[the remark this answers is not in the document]
theryee: gotcha

theryee:
Also noticed that clang has a __has_extension... looking at the documentation it appears that for newer versions of clang extensions are a subset of features: it looks like extensions are clang-specific features but not necessarily in the standard (whereas features are supported by both clang and the standard)... Is it useful to add a EA_COMPILER_HAS_EXTENSION?

Documentation:
https://clang.llvm.org/docs/LanguageExtensions.html

rparolin:

theryee:
Good to go?

PR #188, opened 2018-06-27
Fix unmached optional value constructor/destructor calls

zheland:
Fix unmached optional value constructor/destructor calls count.

Optional assignment operator should use
value copy constructor instead value assigment operator
if target optional does not contain a value.

If source optional does not contain a value,
target value destructor should be called.

### DIFF
--- a/include/EASTL/optional.h
+++ b/include/EASTL/optional.h
@@ -253,13 +253,13 @@ namespace eastl
 		{
 			engaged = other.engaged;
 
-			auto* pOtherStorage = reinterpret_cast<const base_type*>(eastl::addressof(other.val));
-			base_type::operator=(*pOtherStorage);  
+			auto* pOtherValue = reinterpret_cast<const T*>(eastl::addressof(other.val));
+			::new (eastl::addressof(val)) value_type(*pOtherValue);
 		}
 
 	    optional(optional&& other)
 		{
-			eastl::swap(engaged, other.engaged);
+			engaged = other.engaged;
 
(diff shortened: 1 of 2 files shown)


zheland:
The added test cases pass on optional implementation for GCC 7.1.0 and Clang 4.0.0

rparolin:

zheland:
Sorry for the delay in replying.
Conflicts resolved, commit updated.
Modified tests succeed on GCC and Clang ( https://godbolt.org/g/xFdrQC ).

zheland:
From updated commit-message:

Optional copy constructor should use inplace constructor
instead copy assigment operator because constructed value,
does not exist yet and assigment operator can not be used.

Optional move constructor should not use swap operation
for engaged flag because according to C++ draft N4762, page 537:
constexpr optional(optional&& rhs) noexcept(see below );
7 Effects: ... bool(rhs) is unchanged.
8 Ensures: bool(rhs) == bool(*this).

Optional assignment operators should use in-place construct
instead value assignment operator if lhs if nullopt.
If rhs is nullopt it should destruct value in lhs.

Also move assigment operator should not use
swap for engaged flag because according to C++ draft N4762, page 539:
optional<T>& operator=(optional&& rhs) noexcept(see below );
8 Effects: ... The result of the expression bool(rhs) remains unchanged.
10 Ensures: bool(rhs) == bool(*this).

Test where unique ptr moved from one to another optional is
incorrect due to after optional moved bool(optional) remains unchanged.
Test fixed and unique_ptr ownership movement check added.

rparolin:
Thanks for the PR!

PR #189, opened 2018-06-29
Added benchmarks for std::stable_sort.

dave-cope:
The rationale for adding this is the same as for benchmarking std::sort.  Namely that there should be a benchmark against the std implementations of this sort to ensure the EASTL equivalent of std::stable_sort has competitive performance.


rparolin:
Thanks!

PR #192, opened 2018-07-05
Problem with optional constructors

Wawha:
Hello,
there is two problem with eastl::optional constructor.
While nullopt_t has an implicit constructor from int, creating a std::optional from a type which can be implicitly convert to the type store by the optional (and also convert to int), it will prefer the constructor from nullopt_t instead of the right constructor.
This implicit conversion also lead to a bad construction when creating a optional from a volatile int type (it also used the constructor with nullopt_t).
So like the std, I propose to make nullopt_t explicit (and temporary remove test from volatile, while another constructor will need to be implement to handle that case).

There is also another problem in the constructor from const value_type&. The val attribute cannot be constructor directly while it's a storage area. I change the code to use a placement new.

### DIFF
--- a/include/EASTL/optional.h
+++ b/include/EASTL/optional.h
@@ -76,7 +76,7 @@ namespace eastl
 		///////////////////////////////////////////////////////////////////////////////
 		/// optional_storage
 		///
-		template<typename T, bool IsTriviallyDestructible = eastl::is_trivially_destructible_v<T>>
+		template<typename T>
 		struct optional_storage
 		{
 			typedef typename eastl::remove_const<T>::type value_type;
@@ -102,9 +102,11 @@ namespace eastl
 
 			inline ~optional_storage()
 			{
-				if (engaged)
-					destruct_value();
-
+				if(!eastl::is_trivially_destructible_v<T>)
+				{
+					if(engaged)
+						destruct_value();
+				}
 				// engaged = false;  // probably not needed as we are destroying the object
 			}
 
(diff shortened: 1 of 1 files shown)


rparolin:

Wawha:
Hi,
most of the problem I solved previously, was also solved in the last commit for the 3.10.00. So I revert most of my modification.
There is one fix I kept, if the one for copy constructor.
And looking at the code for assignment operator, I see multiples problem (destructor not call, constructor call instead of assignment, engaged set to false after a move, while standard say that it should stay to true)

So I rewrite assignment operators, test it first with _std;;optional<>_, and finally keep test with _eastl:;optional<>_.

Just a quick remark. In my opinion, the specialization for _optional_storage_ class is quite overkill and source of maintenance/bugs, just for the destructor which is different...
At least, we could just specialized a class to manage the destructor, or even simpler and better, just put a _if constexpr_ to know if we have to call the destructor or not. On compiler without _if constexpr_ support, it will lead to the same code, when the destructor is trivial, the code will just go off.

```
inline ~optional_storage()
{
	if(!eastl::is_trivially_destructible_v<T>)
	{
		if (engaged)
			destruct_value();
	}
}
```

Hope it helps. I could repush a commit with that modification if it's ok for you.
Here the commitI create for that modification : https://github.com/Wawha/EASTL/commit/d3ed8ef77b78cdf0ba36187ae979c0cb6edbdc1d

rparolin:
Apologises, but there was another PR that fixed similar issues in eastl::optional.  Can you please re-resolve the code conflicts and re-submit.

Wawha:
Indeed, the other PR solved the same issue. Sorry for the duplication.
Note that I will create a new Pull Request to simplify the implementation of optional_storage like I said previously.

PR #193, opened 2018-07-09
Updating the fixed_pool impl of swap to use Move 

CypherSignal:
While working on some stuff, I noticed that swap() on fixed_vectors with move-only types did not function. This was because the fixed_swap implementation still operated on copy-construction and not move-construction. This update brings the fixed_swap impl back in line to match the functionality in the swap function in utility.h.


rparolin:
Thanks for the PR!

PR #195, opened 2018-07-11
Fix basic_string_view sub-string based basic_string constructor

moretromain:
There is no constructor overload that accepts the current arguments.
This can be easily fixed by using the basic_string_view + allocator constructor that already carries the string size.

### DIFF
--- a/include/EASTL/string.h
+++ b/include/EASTL/string.h
@@ -493,18 +493,8 @@ namespace eastl
 		basic_string(this_type&& x) EA_NOEXCEPT;
 		basic_string(this_type&& x, const allocator_type& allocator);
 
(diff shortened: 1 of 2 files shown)


moretromain:
Ok I’ll see if I can add a relevant test for this, happy to contribute!
I just updated to the last release, and my code won’t compile. I was surprised to see that those additions weren’t covered already.

moretromain:
Done.
First time I mess up with tests, let me know if I did it right. Thanks!

rparolin:
Looks good.  Thanks for the PR.  For future PR's if there is a significant whitespace change please submit a separate PR.  I like to separate whitespace changes and functionality changes in case we need to rollback, etc then we don't lose changes that tidy up the codebase.

Thanks!

PR #196, opened 2018-07-11
C++20: type_trait's endian

theryee:
Doesn't implement support for mixed-endian (AKA middle-endian): this is in contradiction with how the standard describes how mixed-endian support should behave. I'm assuming that mixed-endian support is not useful for the intended audience of EASTL.


rparolin:
Thanks for the PR.

PR #198, opened 2018-07-12
This commit reintroduces two changes that were accidentally lost during

dave-cope:
Potentially other changes were also eliminated during the 3.10.00 release.  This PR just adds back the changes I noticed were missing.


rparolin:
Thanks!  Sorry for the mistake.

PR #201, opened 2018-07-24
Optional: remove useless specialization for optional_storage<>

Wawha:
optional_storage<> have a specialization for type which have a destructor. But all other function/code is the same, so there is too much duplicate code.

We can just use eastl::is_trivially_destructible_v<T> to know if destructor must be called or not.
Due to compiler optimization with compile time condition inside the "if", the exact same code should be generated in release with this implementation.

### DIFF
--- a/include/EASTL/optional.h
+++ b/include/EASTL/optional.h
@@ -76,7 +76,7 @@ namespace eastl
 		///////////////////////////////////////////////////////////////////////////////
 		/// optional_storage
 		///
-		template<typename T, bool IsTriviallyDestructible = eastl::is_trivially_destructible_v<T>>
+		template<typename T>
 		struct optional_storage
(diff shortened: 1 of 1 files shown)


Wawha:
Indeed, it's not the same as if constexpr, but why this implementation, e.g. for native type, you will have a code like this:
			inline ~optional_storage()
			{
				if(false)
				{
					if(engaged)
						destruct_value();
				}
			}

So the compiler will be able to optimize the code and remove the whole code, no? It should generate the same code, in release, as the code with specialization, no?

rparolin:
It disables other optimizations via type inspection for triviality.  That's
why your change broke via the test static asserts.

Wawha:
ok, I close the pull request

PR #202, opened 2018-07-26
Added a new benchmark to better measure the performance of sorting small sub-arrays.

dave-cope:
Rationale: small sub-array performance is important because often efficient algorithms such as merge sort and quick sort will revert to using a simple sorting algorithm for small sub-arrays since they are quite often faster for small arrays.

In addition, the CompareSortPerformance function was slightly modified
so that assignment and comparison counts were recorded for the best
performance run of a test instead of the last run.  Some global variables being referenced were also moved into the body of the function.

### DIFF
--- a/benchmark/source/BenchmarkSort.cpp
+++ b/benchmark/source/BenchmarkSort.cpp
@@ -440,11 +440,6 @@ namespace
 } // namespace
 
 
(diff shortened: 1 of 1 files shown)


rparolin:
Thanks for the PR!

PR #203, opened 2018-08-13
Apply usage of compressed_pair to tuple_vector for empty allocators

CypherSignal:
As suggested here: https://github.com/electronicarts/EASTL/pull/197#discussion_r209438768  This puts the mDataSize member of tuple_vector (used to track how large t_v's allocation itself is) and the mAllocator together into a compressed_pair


rparolin:
Awesome!  Thanks for PR!

Issue #204, opened 2018-08-13
Implement initializer_list support for tuple_vector


CypherSignal:
Expected support should mimic std::vector's interface. That is, the following functions should be added to tuple_vector:

```
tuple_vector<Ts...>(std::initializer_list<value_tuple> iList)
operator=(std::initializer_list<value_tuple> iList)
void assign(std::initializer_list<value_tuple> iList)
const_iterator insert(const_iterator pos, std::initializer_list<value_tuple> iList)
```

CypherSignal:
Work on this is proceeding in https://github.com/CypherSignal/EASTL/tree/tuple_vec-initlist 

A first interface implementation and test-cases are committed already: https://github.com/CypherSignal/EASTL/commit/5facb9b307c2d3292e6ef41f3fbc07edfbf7c818

Besides the functions above, the following have been added as a corollary, since they probably make some sense to have anyway:

```
tuple_vector<Ts...>(const value_tuple* begin, const value_tuple* end)
void assign(const value_tuple* begin, const value_tuple* end)
const_iterator insert(const_iterator pos, const value_tuple* begin, const value_tuple* end)
```

These would allow a developer to more easily insert/assign an array of tuples into tuple_vector instead of doing so one-at-a-time w/ insert or push_back.

There is one issue seen early on, though, with the setup of the initializer_list: because tuple's construction via copy-list-initialization requires explicit use of the constructor, resulting in the init-lists looking like this, more-or-less:

```
			tuple_vector<bool, TestObject, float> testVec;
			tuple<bool, TestObject, float> testTup;

			testTup = tuple<bool, TestObject, float>(true, TestObject(3), 3.0f);
			testVec.insert(testVec.begin(), { 
				tuple<bool, TestObject, float>{true, TestObject(3), 3.0f},
				tuple<bool, TestObject, float>{true, TestObject(3), 3.0f}
				testTup,
				});
```

...that is, for any statically defined tuple, the type of the tuple has to be provided for each initializer. I haven't thought too much about it so far, and maybe there's a way out of this that would involve changing exactly what the initializer_list that is accepted is, but nothing's leaping to mind yet.

CypherSignal:
[...]

PR #205, opened 2018-08-14
Adding initializer_list support to tuple_vector (and support for _conditionally_ explicit support for tuple copy/move ctor)

CypherSignal:
This PR resolves "issue"/task https://github.com/electronicarts/EASTL/issues/204 by adding initializer list support to tuple_vector. This includes the following:

- Implementations for the following functions:
```
tuple_vector<Ts...>(std::initializer_list<value_tuple> iList)
operator=(std::initializer_list<value_tuple> iList)
void assign(std::initializer_list<value_tuple> iList)
const_iterator insert(const_iterator pos, std::initializer_list<value_tuple> iList)
```
- ...which are actually all implemented over the following functions, which may be usable for conversion of, say, vectors of tuples into tuple_vectors 😉
```
tuple_vector<Ts...>(const value_tuple* begin, const value_tuple* end)
void assign(const value_tuple* begin, const value_tuple* end)
const_iterator insert(const_iterator pos, const value_tuple* begin, const value_tuple* end)
```
- Tuple construction is now conditionally explicit (as per a small cpp17 change), i.e. an implicit constructor is utilized if all of the types provided for construction are an exact match for the tuple's definition

Note that this change is independent of https://github.com/electronicarts/EASTL/pull/203 and there *will* be a conflict if a merge is attempted. After this one or https://github.com/electronicarts/EASTL/pull/203 are merged, the other will have to be fixed up due to the new ctors here still referencing the non-compressed-pair version of mAllocator.

### DIFF
--- a/include/EASTL/bonus/fixed_tuple_vector.h
+++ b/include/EASTL/bonus/fixed_tuple_vector.h
@@ -134,7 +134,7 @@ public:
 		const overflow_allocator_type& allocator = EASTL_FIXED_TUPLE_VECTOR_DEFAULT_ALLOCATOR)
 		: base_type(fixed_allocator_type(mBuffer.buffer, allocator), mBuffer.buffer, nodeCount, fixed_allocator_type::kNodeSize)
 	{
-		base_type::DoInitFromTupleArray(iList.begin(), iList.end());
+		base_type::DoInitFromTupleArray(first, last);
 	}
 
 	fixed_tuple_vector(std::initializer_list<typename base_type::value_tuple> iList,
--- a/include/EASTL/bonus/fixed_tuple_vector.h
+++ b/include/EASTL/bonus/fixed_tuple_vector.h
@@ -130,6 +130,20 @@ public:
 		base_type::DoInitFillTuple(n, tup);
 	}
 
+	fixed_tuple_vector(const typename base_type::value_tuple* first, const typename base_type::value_tuple* last,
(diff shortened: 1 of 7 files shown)


CypherSignal:
I'm going to have to come back to and redo the new constructors on tuple. It'll be a couple days until I get around to it.

CypherSignal:
This PR also includes an extra test-case showing an example of code that previously failed without conditionally-explicit ctors: https://github.com/electronicarts/EASTL/blob/96cf7c3c3f405f188eb3b507f22930ce945486f9/test/source/TestTuple.cpp#L451

Also, a few lines below, it has an example block of code that should _continue to fail_ even with conditionally-explicit behaviour, but would work with the incorrect fully implicit behaviour.

rparolin:
Vs2015 still has errors

CypherSignal:
See https://github.com/electronicarts/EASTL/issues/204#issuecomment-414449207 regarding VS2015 errors.

CypherSignal:
Say, will you mind the "doc/EASTL changes.md" file staying in? I think it got in because I based the branch off of the wrong(?) commit for the 3.12.01 release. I can remove it if you want no problem, just trying to figure out what happened to it, and if its presence is intentional or not.

CypherSignal:
Pushing branch with commits all rebased to the current electronicarts:master branch, 98a5a7bd1365668fc7ad792184b97473e0d42ea9 ...

rparolin:
Finally got this across the finish line.  Thank you for your work.

PR #207, opened 2018-08-15
Fix for incorrect or missing platform defines.

ttshughes:
Previous setup for android platform defines incorrectly selects normal arm builds type incorrectly on 64bit builds. Also added missing nnx platform defines.


rparolin:
Sorry.  We don’t have permission from Nintendo to include your submission in the open source EASTL repository at this time.

PR #212, opened 2018-09-11
Fixing problem where eastl::partition on tuple_vector could not compile due to missing swap on eastl::tuple&&

CypherSignal:


CypherSignal:
I'm not sure what the unit-test failure in mingw_64 is about (failing on misc. string activity?). The change should not affect that in any way. Could it be a spurious error that needs a re-run?

rparolin:
Maybe Travis upgraded the default compiler in their image.  That was the cause last time this happened

rparolin:
That's indeed what happened.  The MingGW GCC version went from 7.3.0 -> 8.2.0.
Sorry, I haven't had time lately to look at it.   But it needs to be resolved before I let any PRs in.

rparolin:
Thanks for the PR!  Apologies for the delay in response.

PR #215, opened 2018-09-21
Fixing issue where tuple_vector::operator= for init-list was not used/compiled

CypherSignal:
This was not previously not an issue because the operator= for rvalue-ref of tuple was being invoked implicitly.

This was noticed because tuple_vector::operator=(std::initializer_list<value_tuple> iList) should not have been able to compile. "Assign" for init-list was still being called by other unit-tests, though.


rparolin:

PR #216, opened 2018-09-24
Optimized the implementation of 'merge' to avoid the overhead of calling copy twice (when only one copy has non-zero length).

dave-cope:
This is a fairly minor change, but it gives ~6.5% reduction in time to run the merge_sort/merge_sort_buffer benchmarks on my tests (using MSVC release builds on PC).  The change makes sure that eastl::copy is only invoked once.  Previously copy was called twice even though one of the copies has zero length.  However, this 0 length copy may still have overhead (e.g. calling a non-inline memmove with zero length).

### DIFF
--- a/benchmark/source/BenchmarkAlgorithm.cpp
+++ b/benchmark/source/BenchmarkAlgorithm.cpp
@@ -446,6 +446,23 @@ namespace
 		sprintf(Benchmark::gScratchBuffer, "%p", &*first);
 	}
 
+	template <typename Iterator>
+	void TestMergeStd(EA::StdC::Stopwatch& stopwatch, Iterator firstIn1, Iterator lastIn1, Iterator firstIn2, Iterator lastIn2, Iterator out)
+	{
+		stopwatch.Restart();
+		std::merge(firstIn1, lastIn1, firstIn2, lastIn2, out);
+		stopwatch.Stop();
+		sprintf(Benchmark::gScratchBuffer, "%p", &*out);
+	}
+
+	template <typename Iterator>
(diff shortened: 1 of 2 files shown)


dave-cope:
CI reported one failure, which seems unrelated to the change,

dave-cope:
I have also added some benchmarks to this merge request.  Although the benchmark won't measure the performance impact of the code change (the merge sort benchmark should show improvement though).

benchmark/source/BenchmarkAlgorithm.cpp
```diff
@@ -1106,7 +1123,111 @@ void BenchmarkAlgorithm7(EASTLTest_Rand& /*rng*/, EA::StdC::Stopwatch& stopwatch
...
+		},
+		{
+			"algorithm/merge/vector<TestObject> (unpred)",
+			"algorithm/merge/vector<TestObject> (pred)",
+		},
+	};
+
+	for (int pattern = 0; pattern < 2; pattern++)
+	{
+		if (pattern == 0)
```
rparolin: Consider using an enum or const values express the intent of the pass.

include/EASTL/sort.h
```diff
@@ -191,7 +191,14 @@ namespace eastl
 			++result;
 		}
 
-		return eastl::copy(first2, last2, eastl::copy(first1, last1, result));
+		if (first1 == last1)
```
rparolin: Add an additional comment here so the code doesn't get reverted under future maintenance.

dave-cope: Added!

rparolin:
Thanks for the PR!

PR #221, opened 2018-10-12
system_error.h

theryee:
I noticed that this header was implemented... is this something that you'd want a full implementation for? (This PR only implements the enum class errc)


rparolin:
Hi.  We don't implement this header because it relies on platform specific implementation details.  Unless there is a compelling reason I'd avoid implementing and would just use the implementation provided by the platform.

PR #222, opened 2018-10-12
Improves support for Linux on ARM devices.

zertyz:


rparolin:
FYI, I removed the mention of Raspberry PI (RPI) to avoid confusion that EASTL claims to explicitly support it.  If we wanted to claim that we support RPI we would need to add it to our CI.

Thanks for the PR!

PR #223, opened 2018-10-14
Made optimizations to merge_sort.

dave-cope:
This PR implemented a new variation of merge sort with the ability to call insertion sort for smaller lists as well as optimizations for presorted data.

For random data with inexpensive compare/assignment operations, sort operations typically take only 70-80% of the time.  If data is partially sorted then the performance improvement can be even greater.
For pre-sorted data, this sort is much faster than the previous merge sort implementation.
The new algorithm completes in O(n) time instead of O(n log n) time for sorted data.  For the array sizes in the EASTL benchmarks it is not uncommon to see the sort take only 1-2% of the time as the previous merge sort.

The new algorithm is implemented using a templated class which is currently public.  I could of course put this in an internal namespace if it would be preferable to make the direct use of this class internal.  The only advantage, currently, of allowing access to the class is that users could call the sort with tuned values for InsertionSortLimit.

### DIFF
--- a/include/EASTL/sort.h
+++ b/include/EASTL/sort.h
@@ -442,6 +442,41 @@ namespace eastl
 
 
 
+	namespace Internal
+	{
+		// Sorts a range whose initial (start - first) entries are already sorted.
+		// This function is a useful helper to the tim_sort function.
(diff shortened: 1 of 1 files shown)


rparolin:
Looks great.  Thanks!

PR #226, opened 2018-10-26
Build shared library.

yurivict:


rparolin:
We need some toggle for static vs shared.

rparolin:
Plus this breaks CI

yurivict:
Why do you need the static library?

rparolin:
Feel free to re-open or re-submit the PR when the feedback is addressed.  Thanks.

PR #227, opened 2018-10-26
Fixing ADL manually

jatinx:
This forces the namespace check in eastl namespace for this particular situation.

### DIFF
--- a/include/EASTL/fixed_vector.h
+++ b/include/EASTL/fixed_vector.h
@@ -256,7 +256,7 @@ namespace eastl
 
 		mpBegin = mpEnd = (value_type*)&mBuffer.buffer[0];
 		internalCapacityPtr() = mpBegin + nodeCount;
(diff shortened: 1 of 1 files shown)


rparolin:
Can you add a test case?  Thanks.

jatinx:
I have added test, can you please review it.
The std::variant which needs to be tested with eastl::fixed_vector is only packaged with C++17.

I see the travis.yml file having the c++14 flag.
Can we edit the yml file to include C++17 flag?

test/source/TestFixedVector.cpp
```diff
@@ -533,7 +537,14 @@ int TestFixedVector()
 		EATEST_VERIFY(fv.validate());
 	}
 	#endif
-
+	
+	#if __cplusplus > 201402L
```
rparolin: Looks good, but we have a EA define for CPP17: EA_COMPILER_CPP17_ENABLED.

jatinx:
Added the check of EA_COMPILER_CPP17_ENABLED across test cases.

PR #230, opened 2018-11-05
Adding `get_allocator()` method to `tuple_vector` for parity with other containers

jvidziunas:
Being able to access/copy the allocator has value in some object construction idioms I use and this was missing. Adding an additional `set_allocator()` method would be pretty easy, and I can put that together if we feel there's a need.


rparolin:
Looks good.  Could you add a test?

jvidziunas:
Regular `tuple_vector` should be good to go, but I need to come up with a meaningful solution for the fixed variant. Right now there's no way to customize the overflow allocator, so we'd need to create a new specialization accepting an allocator type. It is possible to write a passing test just with the existing code, but considering all instances compare equal anyway the results aren't very meaningful.

rparolin:
Thanks for the PR!

Issue #233, opened 2018-11-07
macro defining char8_t to char breaks valid code


xaxxon:
The macro in eabase.h breaks valid code:
https://github.com/electronicarts/EASTL/blob/master/test/packages/EABase/include/Common/EABase/eabase.h#L715

```
// this code is fine without the #define
#define char8_t char

enum char8_t : unsigned char {};
```

produces the error:
```

<source>:4:6: error: expected identifier or '{'

enum char8_t : unsigned char {};

     ^
```

https://godbolt.org/z/GjgPra

This example code comes from the popular c++ library {fmt}:

https://github.com/fmtlib/fmt/blob/master/include/fmt/format.h#L404

tahonermann:
To be fair, `fmt` has no more right to the `char8_t` name than EASTL does, so when compiling for C++17 or earlier, this is just a typical name clash across projects.

However, [`char8_t` will be a keyword in C++20](http://wg21.link/p0482).  `fmt` is conditionally defining `char8_t` only if the corresponding feature test macro (`__cpp_char8_t`) is not defined; basically, @vitaut is preparing for C++20 by adding some compatibility features when `char8_t` support is not enabled in the compiler.

EASTL has at least two options for C++20:

1. Change the name of the `char8_t` macro.  This will break source compatibility, but not necessarily binary compatibility, for existing users.
2. Drop the `char8_t` macro in favor of the C++20 `char8_t` type.  This will break binary compatibility, but not necessarily source compatibility (I see EASTL defines `char8_t` as `char`, not as `unsigned char`, so this depends on whether existing code depends on `char8_t` being exactly `char` ).

xaxxon:
What I saw was EASTL doing it globally, FMT doing it in a controlled
fashion... but since EASTL did it globally, FMT being a good citizen
doesn't matter.   Did I see it wrong?

xaxxon:
fmt doesn't use a macro for what it does, though.

tahonermann:
Right, neither EASTL nor fmt is doing anything wrong here.  The issue is that EASTL is defining a macro (as it is allowed to), but that macro is interfering with fmt's use of the same name given to that macro.  Neither project has more or less right to the `char8_t` name.  This is why I stated that this is just a typical name clash across projects.

Note that the comments around the EASTL definition of that macro suggest that the authors didn't actually want to use a macro; they did so (on Apple platforms) to work around a debugger limitation.

vitaut:
[...]

PR #239, opened 2018-12-08
Do not define char8_t if it's defined in a C++20 compiler

christinaa:
Use `__cpp_char8_t` as a feature check to make sure if `char8_t` is
defined by the compiler it is not redefined.


christinaa:
This closes #233 at least for Clang and presumably GCC once it has support for it. MSVC may use something else though, not entirely sure.

Chronial:
Fyi this PR accidentally set the executable permission on `eabase.h`, it's still set on master.

Issue #238, opened 2018-12-07
Test source incompatibility with C++20


aymarino:
A [paper](http://wg21.link/p1008r1) voted into the C++20 draft changes the definition of aggregate to exclude types with any user-declared constructors (previously, the constructors had to be user-provided). Essentially this means any type with an `= delete` or `= default` constructor is no longer an aggregate.

The MSVC team builds EASTL regularly to detect compiler regressions in our live bits, and our implementation exposed this future source issue in test/source/TestOptional.cpp, where a struct wrapping a unique_ptr defines the copy/move ctors as deleted/defaulted respectively, and then aggregate-initializes the ptr inside:

```
struct local
{
    eastl::unique_ptr<int> ptr;

    local(const local&) = delete;
    local(local&&) = default;
    local& operator=(const local&) = delete;
    local& operator=(local&&) = default;
};

eastl::optional<local> o1 = local{eastl::make_unique<int>(42)}; // error: cannot convert from 'initializer list' to 'TestOptional::local'
```

I believe the goal is to prevent copying but allow moving `local`. In this particular case you can actually keep `local` an aggregate without regressing any of those copy/move semantics, because `eastl::unique_ptr` deletes the copy ctor and has a move ctor itself, so `local` would already get that behavior without having to declare the ctors at all. This is the patch we are using to continue testing the library (which is backwards compatible with -std:c++17):

```
diff --git a/test/source/TestOptional.cpp b/test/source/TestOptional.cpp
index f85e15c..c507465 100644
--- a/test/source/TestOptional.cpp
+++ b/test/source/TestOptional.cpp
@@ -542,11 +542,6 @@ int TestOptional()
                        struct local
                        {
                                eastl::unique_ptr<int> ptr;
-
-                               local(const local&) = delete;
-                               local(local&&) = default;
-                               local& operator=(const local&) = delete;
-                               local& operator=(local&&) = default;
                        };

                        eastl::optional<local> o1 = local{eastl::make_unique<int>(42)};
```

I notice EASTL uses -std:c++latest in its builds, which at this point will be including C++20 features. Please note that -std:c++latest is provided as a preview of language features from the latest C++ working draft, and the features are provided as-is without support, and subject to changes or removal as the working draft evolves.

Unless you plant to adopt the bleeding edge of C++20, please consider using -std:c++17 instead, which will be more stable update to update. -std:c++latest is liable to including breaking changes such as this.

[...]

PR #240, opened 2018-12-15
Update test source for C++20 aggregate change

aymarino:
Provides the fix suggested in #238 

The C++20 draft standard changes the definition of aggregate to exclude types which have any user-declared constructor. In this test source, `local` is no longer an aggregate and thus cannot be aggregate-initialized. However, the defaulted and deleted ctors in this struct are redundant, because the `unique_ptr` member already defaults and deletes the same constructors--so, by default the copy ctor is deleted and the move ctor is defaulted.


rparolin:
Thanks for the PR!

PR #244, opened 2018-12-23
Hashtable external key

AntonYudintsev:
In some cases calculation a hash can be expensive and application can already have it calculated.

 This doesn't change any public API functions, but allows external application to save some performance
 by using already caclulated hash in their own hashmap/hashset implementation based on internal EASTL 'hashtable' container

### DIFF
--- a/include/EASTL/internal/hashtable.h
+++ b/include/EASTL/internal/hashtable.h
@@ -1263,10 +1263,20 @@ namespace eastl
 		node_type* DoAllocateNode(value_type&& value);
 		node_type* DoAllocateNode(const value_type& value);
 
-		eastl::pair<iterator, bool> DoInsertKey(true_type, const key_type& key);
-		iterator                    DoInsertKey(false_type, const key_type& key);
-		eastl::pair<iterator, bool> DoInsertKey(true_type, key_type&& key);
-		iterator                    DoInsertKey(false_type, key_type&& key);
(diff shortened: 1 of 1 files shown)


include/EASTL/internal/hashtable.h
```diff
@@ -1263,10 +1263,15 @@ namespace eastl
 		node_type* DoAllocateNode(value_type&& value);
 		node_type* DoAllocateNode(const value_type& value);
 
-		eastl::pair<iterator, bool> DoInsertKey(true_type, const key_type& key);
-		iterator                    DoInsertKey(false_type, const key_type& key);
-		eastl::pair<iterator, bool> DoInsertKey(true_type, key_type&& key);
-		iterator                    DoInsertKey(false_type, key_type&& key);
+		eastl::pair<iterator, bool> DoInsertKey(true_type, const key_type& key, hash_code_t c);
```
rparolin: Add a comment that describes the purpose of these overloads.

AntonYudintsev: just to clarify - comment in code, right?

added overloads mimics current(old) API, so they kept just for compatibility reasons (if someone use it already).

rparolin: Yes.  Comments in code as hints to users who don't know why both sets of API's exist and why they should reach for one or the other.

AntonYudintsev: done

rparolin:
Thanks for the PR!

PR #254, opened 2019-01-08
Constexpr String View

### LINKED ISSUE #250: string_view/wstring_view from string literal
Hello,
It is impossible create a constexpr string_view from a string literal without providing the length. Using MSVC

```c++
struct SomeConstants
{
    // C2131: expression did not evaluate to a constant
    static constexpr const eastl::wstring_view kTest = L"kTest"; 

    // Ok
    static constexpr const eastl::wstring_view kTest = eastl::wstring_view(L"kTest", 5); 
}
```

The reason is calling non-constexpr function _CharStrlen_
```c++
EA_CONSTEXPR basic_string_view(const T* s) : mpBegin(s), mnCount(s != nullptr ? CharStrlen(s) : 0) {}
```

Alan5142:
Using eastl::string_view in a constexpr context causes a compile error, this fixes #250

### DIFF
--- a/include/EASTL/internal/char_traits.h
+++ b/include/EASTL/internal/char_traits.h
@@ -232,19 +232,15 @@ namespace eastl
 		}
 	#endif
 
-	inline size_t CharStrlen(const char8_t* p)
+	inline EA_CONSTEXPR size_t CharStrlen(const char8_t* p)
 	{
-		#if defined(_MSC_VER) || defined(__GNUC__) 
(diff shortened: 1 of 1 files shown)


rparolin:
This fails to compile on VS2015.  Use the 'EA_CPP14_CONSTEXPR' macro is disables constexpr for VS2015 due to its limited implementation.

rparolin:
Looks good.  Thanks for the PR!

PR #257, opened 2019-01-10
Fix an enumeral/non-enumeral type mix

egranata:
With a recent gcc and -Wextra, compiling string.h yields a warning:

EASTL/string.h:3255:55: warning: enumeral and non-enumeral type in conditional expression [-Wextra]
   return (currentCapacity <= SSOLayout::SSO_CAPACITY) ? SSOLayout::SSO_CAPACITY : (2 * currentCapacity);

This patch fixes the warning by means of an explicit cast.

### DIFF
--- a/include/EASTL/string.h
+++ b/include/EASTL/string.h
@@ -337,7 +337,7 @@ namespace eastl
 		// The view of memory when the string data is able to store the string data locally (without a heap allocation).
 		struct SSOLayout
 		{
-			enum : size_type { SSO_CAPACITY = (sizeof(HeapLayout) - sizeof(char)) / sizeof(value_type) };
+			static constexpr size_type SSO_CAPACITY = (sizeof(HeapLayout) - sizeof(char)) / sizeof(value_type);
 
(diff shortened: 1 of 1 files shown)


rparolin:
That's strange because the enum is explicitly typed to size_type.

enum : size_type { SSO_CAPACITY = (sizeof(HeapLayout) - sizeof(char)) / sizeof(value_type) };

egranata:
That was my thinking too, but - alas - gcc begs to differ. I am not enough of a C++ language lawyer to argue whether this is a compiler bug, but that's entirely possible

For reference, here's an example where you can see it happen: https://travis-ci.org/egranata/puppy/jobs/478177540

If you scroll down around line 2800 where userspace starts getting built, you'll see quite a few of these.

rparolin:
Inorder to take away the opportunity for the compiler to pick the wrong type; did you want to convey your PR to change the enum into a variable of size_type instead?   Then the casts shouldn't be necessary.

egranata:
Thanks for your suggestion; let me know if you think the updated CL works!

rparolin:
Thanks for the PR!

PR #268, opened 2019-02-25
Fix unused parameter warning in monostate hash specialization

Meorawr:
The `hash` specialization for monostate had a named parameter (`val`) that wasn't being used by the implementation, and so triggered a C4100 warning under MSVC when compiling with warning level 4 (/W4).

As it's totally unused I just removed the parameter name from the signature rather than using `EA_UNUSED`, for consistency with other PRs (eg. #159).


rparolin:

PR #280, opened 2019-05-29
Fix UB causing crashes with some compilers.

eugeneko:
It's not allowed to call memmove with null pointers even if the size is 0.
In vector::DoInsertValueEnd memmove is called via eastl::uninitialized_move_ptr_if_noexcept unconditionally even for null pointers (when vector is empty).
Therefore, further checks for p in vector::DoFree and allocator::deallocate get optimized away that leads to crashes.

_Maybe_ it's related to this:
https://github.com/electronicarts/EASTL/blob/master/CMakeLists.txt#L58

### DIFF
--- a/include/EASTL/internal/copy_help.h
+++ b/include/EASTL/internal/copy_help.h
@@ -112,6 +112,9 @@ namespace eastl
 		template <typename T>
 		static T* move_or_copy(const T* first, const T* last, T* result)
 		{
+			if (EASTL_UNLIKELY(first == last))
+				return result;
+
 			// We could use memcpy here if there's no range overlap, but memcpy is rarely much faster than memmove.
 			return (T*)memmove(result, first, (size_t)((uintptr_t)last - (uintptr_t)first)) + (last - first);
 		}


rokups:
You forgot assert.

rparolin:
https://en.cppreference.com/w/c/string/byte/memmove

You can pass zero to memmove but it doesn't allow nullptr to be passed.  You change doesn't address passing nullptr to memmove.  Its for zero length requests.

eugeneko:
> You change doesn't address passing nullptr to memmove. Its for zero length requests.

@rparolin, it's legal to pass nullptr into `copy` algorythm **only if** `first == last`, i.e. the empty range `[nullptr, nullptr)` is copied.
If `copy` is called with nullptr and the size is not zero, there's a bug in the caller code.

Therefore, it's enough to check for empty range to prevents well-formed `copy` calls from causing UB.

rparolin:
https://stackoverflow.com/questions/19480609/should-stdcopy-or-stdmove-of-empty-range-require-valid-destination

OK.  Makes sense.  Thanks for the PR.

Issue #293, opened 2019-07-09
std::stack::emplace <> eastl::stack::emplace_back


matt77hias:
In order to be (somewhat more) on par with the std, `eastl::stack::emplace_back` should ideally be renamed to `eastl::stack::emplace` (which is also more consistent to `eastl::stack::push`).

rparolin:
Thanks for reporting this.  Would you like to submit a PR?

matt77hias:
Created a [PR](https://github.com/electronicarts/EASTL/pull/294).

I signed and submitted the CLA as well (_although, technically I am fully employed by Frostbite_).

PR #294, opened 2019-07-10
Renamed eastl::stack::emplace_back to eastl::stack::emplace

matt77hias:
Issue: https://github.com/electronicarts/EASTL/issues/293

### DIFF
--- a/include/EASTL/stack.h
+++ b/include/EASTL/stack.h
@@ -115,7 +115,7 @@ namespace eastl
 		void push(value_type&& x);
 
 		template <class... Args>
-		void emplace_back(Args&&... args);
+		void emplace(Args&&... args);
 
 		void pop();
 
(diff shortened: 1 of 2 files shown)


include/EASTL/stack.h
```diff
@@ -115,7 +115,7 @@ namespace eastl
 		void push(value_type&& x);
 
 		template <class... Args>
-		void emplace_back(Args&&... args);
+		void emplace(Args&&... args);
```
rparolin: We should maintain backwards compat and give users an opportunity to migrate off the old API.

matt77hias: Does EASTL target C++14 as minimum C++ standard. If so we can add `[[deprecated]]` to the old one to make it explicitly?

rparolin: Yes.  We have EA_DEPRECATED portability macros to enable/disable that... or use other features of compiler if that doesn't work.

PR #298, opened 2019-07-17
 Replaced queue|stack::emplace_back with queue|stack::emplace.

matt77hias:


rparolin:
I fixed this internally.  It will be included in the next EASTL release.

PR #303, opened 2019-07-31
Fix memory leak in string_hash_map::insert_or_assign

theryee:
Previous implementation always duplicated the key
(even if a previous key already existed).

A leak occurred because the clear was not aware of the
previous key that was discared by the newly duplicated key.

This solution only duplicates a new key if necessary.

### DIFF
--- a/test/source/TestStringHashMap.cpp
+++ b/test/source/TestStringHashMap.cpp
@@ -4,6 +4,7 @@
 
 
 #include "EASTLTest.h"
+#include <EASTL/fixed_allocator.h>
 #include <EASTL/string_hash_map.h>
 #include <EAStdC/EAString.h>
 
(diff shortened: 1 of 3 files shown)


rparolin:
Ah, good catch.

Can you write a test that catches this regression?  I've written tracking allocators in previous tests that can be used to ensure all allocations are freed at scope exit.

theryee:
I based it on the example I found in TestFixedMap.cpp

theryee:
I also tested without the fix: 4 objects were allocated.. which makes sense because of the leaking key.

theryee:
Hmm... seems to be triggering this assert one msvc.

```
			// Ensure the result is correctly aligned.  An assertion likely indicates a mismatch between EASTL_ALLOCATOR_MIN_ALIGNMENT and the minimum alignment
			// of EASTLAlloc.  If there is a mismatch it may be necessary to define EASTL_ALLOCATOR_MIN_ALIGNMENT to be the minimum alignment of EASTLAlloc, or
			// to increase the alignment of EASTLAlloc to match EASTL_ALLOCATOR_MIN_ALIGNMENT.
			EASTL_ASSERT((reinterpret_cast<size_t>(result)& ~(alignment - 1)) == reinterpret_cast<size_t>(result));
```

test/source/TestStringHashMap.cpp
```diff
@@ -278,6 +279,34 @@ int TestStringHashMap()
 
 			EATEST_VERIFY(m.size() == 1);
 		}
+
+		{
+			typedef string_hash_map<TestObject, hash<const char*>, str_equal_to<const char*>, fixed_allocator> fixed_string_hash_map;
```
rparolin:
```///////////////////////////////////////////////////////////////////////////////
/// CountingAllocator
///
/// Counts allocation events allowing unit tests to validate assumptions.
///
class CountingAllocator : public eastl::allocator
```

I was thinking of the above test allocator.

```	// Check for memory leaks by using the 'CountingAllocator' to ensure no active allocation after tests have completed.
	CountingAllocator::resetCount();
	nErrorCount += TestBasicString<eastl::basic_string<char, CountingAllocator>>();
	VERIFY(CountingAllocator::getActiveAllocationCount() == 0); 
```

rparolin:
Thanks for the PR!

PR #311, opened 2019-08-17
Fix unused parameter warnings in a DecodePart overload

hamzasood:
This fixes a function that triggers C4100 (unreferenced formal parameter) when compiling with MSVC.


rparolin:
Thanks for the PR.

PR #316, opened 2019-10-01
Added empty base optimization wrt to allocator to list and slist.

wmamrak:


rparolin:
Same here.  Please add a unit test to validate EBO behaviour.

rparolin:
Thanks for the PR.

PR #329, opened 2019-12-05
Avoid non-public header thr/xtimec.h on Win32

CaseyCarter:
...which doesn't seem to be used in any case.

Note that the name and location of this internal header are subject to change without notice, as happened in the Visual Studio 2019 version 16.4 release. *Not* including the header seems to avoid the problem quite nicely ;)

### DIFF
--- a/include/EASTL/chrono.h
+++ b/include/EASTL/chrono.h
@@ -43,6 +43,10 @@
 	#endif
 	EA_RESTORE_ALL_VC_WARNINGS()
 	#pragma warning(pop)
+#endif
+
+#if defined(EA_PLATFORM_MICROSOFT) && !defined(EA_PLATFORM_MINGW)
+	// Nothing to do
 #elif defined(EA_PLATFORM_SONY)
 	#include <Dinkum/threads/xtimec.h>
 	#include <kernel.h>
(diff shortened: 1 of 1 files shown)


include/EASTL/chrono.h
```diff
@@ -43,10 +43,6 @@
 	#endif
 	EA_RESTORE_ALL_VC_WARNINGS()
 	#pragma warning(pop)
-#endif
```
StephanTLavavej: The intent of this PR appears to be correct (avoid including `<thr/xtimec.h>` for MSVC), but this preprocessor coalescing appears to be incorrect. Note how left line 48 indicates that MICROSOFT and MINGW can be defined simultaneously. Previously, MICROSOFT and MINGW would activate line 29, and separately would skip left line 48 and activate left line 55. Now, the coalescing means that line 29 will absorb the MICROSOFT and MINGW scenario.

I believe that this PR should be changed to be non-intrusive, by simply replacing the `<thr/xtimec.h>` include with a do-nothing comment, and not coalescing the preprocessor logic.

CaseyCarter: I agree; presumably CI didn't catch this due to the absence of MinGW coverage.

rparolin:
Excellent.  Thanks for the PR!

PR #330, opened 2019-12-05
Fix move flavour of vector::DoAssign[FromIterator]

virt00l:
Expression
>bMove ? should_move_tag() : should_copy_tag()

wasn't working as expected, basically always coalescenig as
should_copy_tag type (*) which resulted always copy-constructed
code-path to be taken instead of move-contructed's one regardless of
bMove value. Which is perfomance degradation

This code-path e.g. used in fixed_vector's move constructor

(*) You can read more about underlying reasons for this behaviour in
https://stackoverflow.com/questions/8535226/return-type-of-ternary-conditional-operator

### DIFF
--- a/include/EASTL/vector.h
+++ b/include/EASTL/vector.h
@@ -322,7 +322,9 @@ namespace eastl
 		// (iterator categories). This is because in these cases there is an optimized
 		// implementation that can be had for some cases relative to others. Functions
 		// which aren't referenced are neither compiled nor linked into the application.
-		struct should_copy_tag{}; struct should_move_tag : public should_copy_tag{};
+		template <bool bMove> struct should_move_or_copy_tag{};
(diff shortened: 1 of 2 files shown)


rparolin:
Awesome.  Can you write a test for this change?

virt00l:
I was wrong about non-working fixed_vector's ctor/assignment. It do work, except one specific code path/method:
>fixed_vector(this_type&& x, const overflow_allocator_type& overflowAllocator);
where it (incorrectly) do copying instead of movement. Now it works too. 

Also new code path slightly less odd - DoAssignFromIterator( ) with bMove=true calls 
DoRealloc() with proper tag (should_move_tag), altough eastl::uninitialized_copy_ptr() for move_iterator<generic_iterator<>> do movement, instead of copying.

rparolin:
Awesome.  Thanks for the PR!

PR #331, opened 2019-12-05
Fix vector::push_back_uninitialized() growth strategy

virt00l:
It was capacity()+1 by some reason, which is extremely slow for certain
usage patterns. Change it's to default GetNewCapacity() which is used
e.g. for normal push_back


rparolin:
I can't integrate this fix back as it breaks existing custom behaviour not mandated by the standard.   Sorry about that.  I'd be open to an overload or a different function which enables the growth behaviour.

PR #332, opened 2019-12-05
Fix back_insert_iterator move flavour

virt00l:
Add push_back move (rvalue) flavour in order to code like
eastl::move(container.begin(), container.end(), eastl::back_inserter(container))
to work as intended (which is to use move-version of push_back obviously)

### DIFF
--- a/include/EASTL/iterator.h
+++ b/include/EASTL/iterator.h
@@ -639,6 +639,9 @@ namespace eastl
 		back_insert_iterator& operator=(const_reference value)
 			{ container.push_back(value); return *this; }
 
+		back_insert_iterator& operator=(typename Container::value_type&& value)
+			{ container.push_back(eastl::move(value)); return *this; }
+
 		back_insert_iterator& operator*()
 			{ return *this; }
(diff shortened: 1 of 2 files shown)


include/EASTL/iterator.h
```diff
@@ -639,6 +639,9 @@ namespace eastl
 		back_insert_iterator& operator=(const_reference value)
 			{ container.push_back(value); return *this; }
 
+		back_insert_iterator& operator=(typename Container::value_type && value)
```
rparolin: Please align the move-semantic reference with value_type.

rparolin:
Also, please write a test for the new overload.

include/EASTL/iterator.h
virt00l: Do you mean something like
>back_insert_iterator& operator=(typename Container::value_type&& value)
?

rparolin: Yes please.

rparolin:
Thanks for the PR!

PR #333, opened 2019-12-09
Using of qualified eastl compare function names

miherius:
Fixing of compilation error when some of `eastl` containers used with `std` types. Reason: ambiguity of `equal()` and `lexicographical_compare()` functions call in overloaded compare operators, which conflicts with its `std` counterparts. So, you can't compare `eastl::vector<std::string>` using `operators==()`, but compare of `eastl::array<std::string>` works fine, because latter use qualified names for its compare operators.

This PR just unifies using of compare functions across all `eastl` containers. So, all unqualified calls to `equal()` and `lexicographical_compare()` functions replaced with quailified ones.

I'm not sure that's a good idea to write tests which can't be compiled in the pre-PR version, so I skipped this part.

Tested on VS 2017 and GCC 7.4

### DIFF
--- a/include/EASTL/bitvector.h
+++ b/include/EASTL/bitvector.h
@@ -1417,7 +1417,7 @@ namespace eastl
 						   const bitvector<Allocator, Element, Container>& b)
 	{
 		// To do: Replace this with a smart compare implementation. This is much slower than it needs to be.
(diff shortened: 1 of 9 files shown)


rparolin:
Thanks for the PR!

PR #335, opened 2019-12-25
Fix string_view::find_first_of buffer overflow bug

nemerle:
The `find_first_of` overload taking `string_view` parameter was passing the size of `this`, instead of the `sw`'s size, which caused incorrect read of source view's memory.


rparolin:
Thanks for the PR!

PR #336, opened 2020-01-13
fixed atomic_exchange_explicit() return with shared_ptr<>

Wawha:
atomic_exchange_explicit() for shared_ptr<> doesn't have a return.


rparolin:
Cool.  Thanks for the PR!

PR #337, opened 2020-01-16
Add vcpkg installation instructions

NancyLi1013:
EASTL is available as a port in [vcpkg](https://github.com/microsoft/vcpkg), a C++ library manager that simplifies installation for EASTL and other project dependencies. Documenting the install process here will help users get started by providing a single set of commands to build EASTL , ready to be included in their projects.

We also test whether our library ports build in various configurations (dynamic, static) on various platforms (OSX, Linux, Windows: x86, x64, UWP) to keep a wide coverage for users.

I'm a maintainer for vcpkg, and [here is what the port script looks like](https://github.com/microsoft/vcpkg/blob/master/ports/eastl/portfile.cmake). We try to keep the library maintained as close as possible to the original library.


rparolin:
Awesome!  Thanks for making EASTL available via vcpkg. Thanks for the PR!

PR #338, opened 2020-01-23
Added span::front(), back(), and subspan()

Meorawr:
Implements the missing `span::front()` and `span::back()` element access members, as well as the `span::subspan()` operation.

Each of the new members follows the same implementation rules used by the existing `span` code and will assert if the user would otherwise invoke undefined behaviour:
* `span::front()` and `span::back()` will assert if the span is empty.
* `span::subspan()` will assert if offset or count exceed the bounds of the span.

Tests are included for each member as follows:
* `span::front()` and `span::back()` are tested along with the existing subscript operator case which I've renamed to TestElementAccess; didn't seem worth putting these in a separate test but calling them subscript operators wasn't really accurate.
* `span::subspan()` tests cases where the range of the subspan is equal to the full range of the original span, a subrange, and an offset with the default count parameter (`dynamic_extent`).

Tested successfully under Clang 9 and GCC 9.2.0.

### DIFF
--- a/include/EASTL/span.h
+++ b/include/EASTL/span.h
@@ -400,7 +400,7 @@ namespace eastl
 	span<T, Extent>::subspan() const
 	{
 		EASTL_ASSERT_MSG(bounds_check(Offset),                                  "undefined behaviour accessing out of bounds");
-		EASTL_ASSERT_MSG(Count == dynamic_extent || Count <= (size() - Offset), "undefined behaviour accessing out of bounds");
+		EASTL_ASSERT_MSG(Count == dynamic_extent || Count <= (size() - Offset), "undefined behaviour exceeding size of span");
 
 		return {data() + Offset, Count == dynamic_extent ? size() - Offset : Count};
 	}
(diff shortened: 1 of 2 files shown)


include/EASTL/span.h
```diff
@@ -41,9 +43,18 @@ namespace eastl
...
+		// SubspanExtent
+		//
+		// Integral constant that calculates the resulting extent of a templated subspan operation.
+		//
+		//   If Count is not dynamic_extent then SubspanExtent::value is Count,
+		//   otherwise, if Extent is not dynamic_extent, SubspanExtent::value is (Extent - Offset),
+		//   otherwise, SubspanExtent::value is dynamic_extent.
+		//
+		template<size_t Extent, size_t Offset, size_t Count>
+		struct SubspanExtent : eastl::integral_constant<size_t, (Count != dynamic_extent ? Count : (Extent != dynamic_extent ? (Extent - Offset) : dynamic_extent))> {};
```
Meorawr: Wasn't too sure on how to best deal with the rather lengthy subspan extent expression.

I didn't like duplicating it across its two usage locations, so I put it into the Internal namespace as an integral_constant as it felt like a safe option in terms of compiler support.

If you'd rather just be inlined into its two usage locations (or handled another way) then I'll happily accommodate.

rparolin:
Looks great!  Thanks for doing this work.  Much appreciated.

PR #341, opened 2020-01-24
Fix missing explicit constructors in span::rbegin/rend()

### LINKED ISSUE #339: Span rbegin/rend fail to compile (VS2019)
Hello!

It seems eastl::span rbegin() and rend() do not call reverse_iterator's constuctor explicity, and unforunately this constructor is marked explicit.

I'm not sure why the tests for span are not catching that? rbegin and rend are used there.

https://github.com/electronicarts/EASTL/blob/1774cb4839a357a2f3a1cc58ec54575a8158b055/include/EASTL/span.h#L312-L316

Meorawr:
Fixes #339. The ticket said that rbegin/rend had tests, but this wasn't the case so I've added them as well.


rparolin:
Thanks for the PR!

PR #342, opened 2020-01-24
Fix overriding of -std=c++17 flags

Meorawr:
From local testing with other changes I'd noticed that C++17 features like deduction guides weren't being properly enabled in tests.

The logic of the CommonCppFlags script causes the flags set for C++17 support to be overridden by C++14 flags instead due to the if() check immediately after in each branch being a GREATER_OR_EQUALS check; for example the logic for Clang would first set `-std=c++17` for Clang 5+, and then the next check would append `-std=c++14` for Clang 3.5+.

I've added additional checks on the conditions to match the other ones present for each compiler to prevent this.

Note that a better way of resolving this might be to take advantage of the [`CMAKE_CXX_STANDARD` variable](https://cmake.org/cmake/help/v3.16/prop_tgt/CXX_STANDARD.html#prop_tgt:CXX_STANDARD), which makes CMake automatically add the proper `-std=<standard>` flag for a specified standard version based on what it knows the compiler supports. If the compiler doesn't support the requested standard, it'll decay it downwards if the accompanying `CXX_STANDARD_REQUIRED` flag is left unset or disabled. This could replace basically all of the checks in the file, but I'm unsure if it'd introduce any issues:

```cmake
if(NOT (CMAKE_VERSION VERSION_LESS 3.8))
     # CMake 3.8+ supports flags for C++17.
     set(CMAKE_CXX_STANDARD 17)
else()
     # CMake 3.1+ (required by the top-level listfile) supports flags for C++14.
     set(CMAKE_CXX_STANDARD 14)
endif()
```


Meorawr:
[Original build failure](https://travis-ci.org/electronicarts/EASTL/jobs/641440200#L220)) seems to stem from `std::random_shuffle` having been deprecated and then removed as of C++17, and presumably Clang under the OSX follows that strictly.

The benchmark's only reference was commented out, so I've just deleted it entirely to see if that'll build successfully. Open to any suggestions on better ways to handle it if removing the unused benchmark isn't palatable.

[Next build failure](https://travis-ci.org/electronicarts/EASTL/jobs/641449033#L245) seems to then be that the OSX environment (Xcode 9.4) is C++17-compatible from the compiler's perspective, but libc++ lacked an implementation of `<variant>`. As `__has_include` is part of the C++17 standard I'm hoping that can be used reliably to  test for its existence as an alternative.

rparolin:
Thanks for the PR!

PR #343, opened 2020-01-24
Add array deduction guides

Meorawr:
The guides are defined as part of the C++17 standard, and are gated behind a check for the `__cpp_deduction_guides` define as per the span implementation.

Note that the tests won't actually be executed unless #342 is merged in (except under MSVC), since the issue that PR solves prevents GCC/Clang from compiling the tests as anything newer than C++14.


rparolin:
Thanks for the PR!

PR #344, opened 2020-01-24
Fix GCC 9.2 constexpr compile errors in array tests

Meorawr:
Attempted fix for the issue noted in #325 with GCC 9.2.0. Compiling the array tests causes GCC to choke on the subscript operator of reverse_iterator inside of a constant expression.

Using the (commented out) `-n - 1` subscript expression seems to work around this. I can't see anything in the repository history for why it was commented out to begin with, so there might be some implication I'm unaware of, but the tests all still pass under GCC 9.2.0 and Clang 9 for myself at least.


rparolin:
Thanks for the PR!

PR #345, opened 2020-01-25
C++20 char8_t string/string_view support improvements

Meorawr:
* Added `operator ""sv` support for creating a `string_view` from a `char8_t` string literal. Additionally added the `operator ""_sv`to enable testing.
* Added `operator ""s` support for creating a `string` from a `char8_t` string literal.
* Added specializations of `eastl::hash` for both `u8string` and `u8string_view`.

These are only enabled when `char8_t` is a distinct type under C++20, since otherwise the existing `char`-based definitions are sufficient.

* Added test variants for `u8string` and `u8string_view`.
* Added support for some commented-out tests around `wchar_t` support with strings.

Unfortunately there's a number of errors from EAStdC and some other submodules when compiling the full test suite in C++20 mode due to lots of structures and signatures mixing `char` and `char8_t`, so the results from these wouldn't show in CI (even under MSVC, since [the current setup doesn't support the `/Zc:char8_t` flag](https://travis-ci.org/electronicarts/EASTL/jobs/641814300#L110)).

I've added a build option (`EASTL_BUILD_CHAR8T_TESTS`) which will attempt to build just the string/string_view tests with `char8_t` support if CMake is happy that the compiler supports either the `-fchar8_t` or `/Zc:char8_t` flags. As the Clang compiler in the Travis environment is version 7, which supposedly supports it, I've added a matrix to the CI setup to enable this (and if it works, it should show up in the build log). Locally I've verified all the tests pass with both GCC 9.2.0 and Clang 9 with this option both set and unset.

I've had to add a few additional overloads for DecodePart to handle string test cases where `append_convert` would be called with the same source and destination character types for `char8_t` and `wchar_t`, or where the destination buffer was `char8_t`-based. These just cast the parameters to `char*`s and forward to the appropriate overloads.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,9 +34,6 @@ matrix:
         - compiler: clang "release build with clang to trigger MOJI check"
           env: EASTL_CONFIG=Release USE_MOJI_CHECK=yes
           os: linux
-        - compiler: clang # C++20 char8_t support is present in Clang 7+.
-          env: EASTL_CONFIG=Release EASTL_BUILD_CHAR8T_TESTS=ON
-          os: linux
         - compiler: msvc
           env: EASTL_CONFIG=Release CXXFLAGS="/std:c++latest /Zc:char8_t"
           os: windows 
@@ -72,7 +69,7 @@ install:
 before_script:
   - mkdir build_$EASTL_CONFIG
   - cd build_$EASTL_CONFIG
-  - cmake .. -DEASTL_BUILD_BENCHMARK:BOOL=ON -DEASTL_BUILD_TESTS:BOOL=ON -DEASTL_BUILD_CHAR8T_TESTS:BOOL=${EASTL_BUILD_CHAR8T_TESTS}
+  - cmake .. -DEASTL_BUILD_BENCHMARK:BOOL=ON -DEASTL_BUILD_TESTS:BOOL=ON
   - cmake --build . --config $EASTL_CONFIG
 
 script:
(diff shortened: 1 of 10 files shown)


Meorawr:
The new Clang CI matrix entry and CMake option [worked in proving `-fchar8_t` is supported](https://travis-ci.org/electronicarts/EASTL/jobs/641852270#L511) in the Travis environment.

The latest changeset I've just pushed adds a definition to the test compilation which, if set but `EA_CHAR8_UNIQUE` evaluates to false, should trigger a compile-time error. This should hopefully be sufficient in proving that the CI is successfully building and executing the string/string_view tests with char8_t support under Clang 7.

.travis.yml
```diff
@@ -69,7 +72,7 @@ install:
 before_script:
   - mkdir build_$EASTL_CONFIG
   - cd build_$EASTL_CONFIG
-  - cmake .. -DEASTL_BUILD_BENCHMARK:BOOL=ON -DEASTL_BUILD_TESTS:BOOL=ON
+  - cmake .. -DEASTL_BUILD_BENCHMARK:BOOL=ON -DEASTL_BUILD_TESTS:BOOL=ON -DEASTL_BUILD_CHAR8T_TESTS:BOOL=${EASTL_BUILD_CHAR8T_TESTS}
```
rparolin: I would just enable the tests if the compiler is configured to enabled char8_t.

rparolin:
Awesome!  Thanks for the PR!

PR #348, opened 2020-02-10
add enable_shared_from_this::weak_from_this() present since C++17

Wawha:
Add enable_shared_from_this::weak_from_this() because it's now available since C++17.
cf. https://en.cppreference.com/w/cpp/memory/enable_shared_from_this/weak_from_this


rparolin:
Awesome. Thanks for the PR!

PR #349, opened 2020-02-11
Add Conan to package managers

uilianries:
Add description of Conan package manager and EASTL availability there.

/cc @danimtb


rparolin:
Cool.  Thanks!

PR #350, opened 2020-02-25
Fix circular dependency in fixed_function.h

### LINKED ISSUE #275: Circular header dependency
The two headers `internal/function.h` and `internal/function_detail.h` end up including each other.

The cycle is:
1. `<EASTL/internal/function_detail.h>`
2. `<EASTL/utility.h>`
3. `<EASTL/functional.h>`
4. `<EASTL/internal/function.h>`
5. `<EASTL/internal/function_detail.h>`

This simple one line program fails to compile with lots of errors:

    #include <EASTL/fixed_function.h>

If `internal/function_detail.h` is first included from any file other than `internal/function.h` (e.g. from `fixed_function.h`), then the build will fail.

Meorawr:
Fixes #275.

The referenced issue notes a compilation error that occurs when `fixed_function.h` is included ahead of `functional.h`. This can be demonstrated by just compiling a lone source file with the following:
```cpp
#include <EASTL/fixed_function.h>
```

This results in the following (shortened) inclusion order as dumped via MSVC's `/showIncludes` option, and error:
```
include\EASTL\fixed_function.h
 include\EASTL\internal\function_detail.h      # function_detail from fixed_function...
  include\EASTL\utility.h                               
   include\EASTL\functional.h                  # utility requires reference_wrapper traits...
    include\EASTL\internal\mem_fn.h
    include\EASTL\internal\function.h
     include\EASTL\internal\function_detail.h  # Oops!
include\EASTL\internal\function.h(31): error C2143: syntax error: missing ',' before '<'
include\EASTL\internal\function.h(127): note: see reference to class template instantiation 'eastl::function<R(Args...)>' being compiled
include\EASTL\internal\function.h(34): error C2061: syntax error: identifier 'function_detail'
include\EASTL\internal\function.h(36): error C2653: 'Base': is not a class or namespace name
include\EASTL\internal\function.h(36): error C2146: syntax error: missing ';' before identifier 'result_type'
```

This doesn't show up in the test suite as-is since `functional.h` is included (via `EASTLTest.h` and `algorithm.h`) prior to `fixed_function.h`, so the first commit reorders those includes to try and prevent this issue going by undetected.

As `utility.h` only requires the `eastl::reference_wrapper` type traits, I've opted to fix this by just pulling in `functional_base.h` (instead of `functional.h`).

This will affect users if they're expecting the full breadth of `functional.h` to have been made available, so an alternative could be instead to just make `fixed_function.h` include `functional.h` (instead of `function_detail.h`) at the expense of making that header unnecessarily large.

### DIFF
--- a/test/source/TestFixedFunction.cpp
+++ b/test/source/TestFixedFunction.cpp
@@ -2,10 +2,11 @@
 // Copyright (c) Electronic Arts Inc. All rights reserved.
 /////////////////////////////////////////////////////////////////////////////
 
+#include <EASTL/fixed_function.h>
+
 #include <EABase/eabase.h>
 #include <EAAssert/eaassert.h>
 #include "EASTLTest.h"
-#include <EASTL/fixed_function.h>
 #include <EASTL/numeric.h>
 
 EA_DISABLE_ALL_VC_WARNINGS()
--- a/test/source/TestFixedFunction.cpp
+++ b/test/source/TestFixedFunction.cpp
@@ -2,10 +2,13 @@
 // Copyright (c) Electronic Arts Inc. All rights reserved.
 /////////////////////////////////////////////////////////////////////////////
 
-#include <EASTL/fixed_function.h>
-
 #include <EABase/eabase.h>
 #include <EAAssert/eaassert.h>
+
+// Included prior to EASTLTest.h to guard against the following bug resurfacing:
+// https://github.com/electronicarts/EASTL/issues/275
+#include <EASTL/fixed_function.h>
+
 #include "EASTLTest.h"
 #include <EASTL/numeric.h>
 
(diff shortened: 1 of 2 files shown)


Meorawr:
Just to note, failure of the first commit via the CI was intended so to [demonstrate the issue](https://travis-ci.org/electronicarts/EASTL/jobs/655005250#L253) and the effect the patch has on fixing it.

test/source/TestFixedFunction.cpp
```diff
@@ -2,10 +2,11 @@
 // Copyright (c) Electronic Arts Inc. All rights reserved.
 /////////////////////////////////////////////////////////////////////////////
 
+#include <EASTL/fixed_function.h>
```
rparolin: Is this change required?  EABase.h should be fine as the first header.

Meorawr: Moved it below `eabase.h` and `eaassert.h`, and threw in a comment to explain why it's before `EASTLTest.h` so the reasoning is clearer for future readers of that file.

rparolin:

PR #358, opened 2020-04-09
fix bitvector iterators

AntonYudintsev:
use mContainer.begin() instead of &mContainer[0],
since the latter causes out of bounds exception in non release build


rparolin:
Can we add a test that highlights a bug for this?

AntonYudintsev:
Any comments why not merge it?

rparolin:
Sorry nothing to do with your change but the Travis CI integration seems to have been broken.  I'm trying to get cycles to look into it.  Stay tuned.

PR #359, opened 2020-04-12
Enable compilation without building tests

wmamrak:


rparolin:
Why is this necessary?

wmamrak:
EABase target is not defined, thus EASTL target lacks `EASTL\test\packages\EABase\include\Common` in its include directories.

rparolin:
FYI - triggering a CI build.

PR #373, opened 2020-06-08
fixed unique_ptr::reset() instructions order. Internal pointer must be updated before calling deleting object

Wawha:
Currently, unique_ptr::reset(), delete the object (if present) and then change the internal pointer.
But, according to standard (see https://en.cppreference.com/w/cpp/memory/unique_ptr/reset), the operation must be done in this order : 
* Saves a copy of the current pointer old_ptr = current_ptr
* Overwrites the current pointer with the argument current_ptr = ptr
* If the old pointer was non-empty, deletes the previously managed object if(old_ptr) get_deleter()(old_ptr).

It make sense, in order to avoid having a internal pointer on an object that will be destroyed.

### DIFF
--- a/include/EASTL/unique_ptr.h
+++ b/include/EASTL/unique_ptr.h
@@ -210,10 +210,8 @@ namespace eastl
 		{
 			if (pValue != mPair.first())
 			{
-				if (mPair.first())
-					get_deleter()(mPair.first());
-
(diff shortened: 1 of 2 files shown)


rparolin:
Nice catch!  Thanks for the PR.

PR #378, opened 2020-06-20
eastl::shared_ptr<>::reset(): function not thread safe

Wawha:
reset() function of shared_ptr is not thread safe.
It takes so many times to understand/find where come from that memory leak I have in my code. The problem was so difficult to reproduce and isolate!

So here the code:
```
	inline void ref_count_sp::release()
	{
		EASTL_ASSERT((mRefCount > 0) && (mWeakRefCount > 0));
		if(Internal::atomic_decrement(&mRefCount) > 0)
			Internal::atomic_decrement(&mWeakRefCount);
		else
		{
			free_value();

			if(Internal::atomic_decrement(&mWeakRefCount) == 0)
				free_ref_count_sp();
		}
	}
```

If 2 threads, with 2 shared_ptr<> pointed the same value, and are calling the release() function at the same time, the `free_ref_count_sp();` could be not called.
It happens, when thread 1 is calling `Internal::atomic_decrement(&mWeakRefCount);` (due `m_RefCount > 0`) after the thread 2 call the code `if(Internal::atomic_decrement(&mWeakRefCount) == 0)`. In that case `free_ref_count_sp();` is not called, because on the thread 2, `mWeakRefCount` will have value 1.

The fix consist to remove that else condition and process each counter independently : 
```
	inline void ref_count_sp::release()
	{
		EASTL_ASSERT((mRefCount > 0) && (mWeakRefCount > 0));
		if(Internal::atomic_decrement(&mRefCount) == 0)
			free_value();

		if(Internal::atomic_decrement(&mWeakRefCount) == 0)
			free_ref_count_sp();
	}
```

I try to write UTest for that case. For the moment, I used std::future and std::async because it's easier for me. If I have to used EAThread, could you tell me the best eastl function to used instead.

### DIFF
--- a/include/EASTL/shared_ptr.h
+++ b/include/EASTL/shared_ptr.h
@@ -162,15 +162,11 @@ namespace eastl
 	inline void ref_count_sp::release()
 	{
 		EASTL_ASSERT((mRefCount > 0) && (mWeakRefCount > 0));
-		if(Internal::atomic_decrement(&mRefCount) > 0)
-			Internal::atomic_decrement(&mWeakRefCount);
-		else
-		{
+		if(Internal::atomic_decrement(&mRefCount) == 0)
 			free_value();
 
-			if(Internal::atomic_decrement(&mWeakRefCount) == 0)
-				free_ref_count_sp();
-		}
+		if(Internal::atomic_decrement(&mWeakRefCount) == 0)
+			free_ref_count_sp();
 	}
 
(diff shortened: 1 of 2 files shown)


MaxEWinkler:
Thank you so much for the fix.
This is a really good find and a definite issue.
Thanks again for taking the time to figure this out and sending us the patch :).

PR #380, opened 2020-07-03
Fix GCC 9 string SSO segfault

virt00l:
Old implementation confused GCC 9 (reproduced on 9.1.1 & 9.3.1, not
reproduced on earlier versions) on string assigment of
exact SSO_CAPACITY (23) length. It caused
SSO string to be non-terminated and further segfault on attempt
to interpret actual string content as pointers data (i.e. heap string)

New propsed implementation works with all major compilers including
GCC, MSVC & Clang (tested both normal & sanitized builds in fairly
large commertial project)

Note on implementation:
* Don't memset whole string in default ctor anymore (as for empty
initialized SSO string we don't need to change anything besides
first null-terminated character & mnRemainingSize) - this should
save a few cycles on default string ctor which is quite frequently
executing code
* AllocateSelf(size_type n) now sets size of string to passed size
(instead of 0 as it was before), which is usefull in much more
cases then old behaviour (most of ctors calls {Size,Range}Initialize()
which calls this method). This also reduces number of overwrites for
SSO string size, which is also should be win for not-so-smart compilers
that lacks (or have poor) "dead store elimination" optimization

Change-Id: I2618f7be4777760d45f8094300694d173d4ce332

### DIFF
--- a/include/EASTL/string.h
+++ b/include/EASTL/string.h
@@ -389,7 +389,7 @@ namespace eastl
 				RawLayout raw;
 			};
 
-			Layout()                                                  { ResetToSSO(); SetSSOSize(0); } // start as SSO by default
+			Layout()                                                  { ResetToSSO(); } // start as SSO by default
 			Layout(const Layout& other)                               { Copy(*this, other); }
 			Layout(Layout&& other)                                    { Move(*this, other); }
 			Layout& operator=(const Layout& other)                    { Copy(*this, other); return *this; }
@@ -481,7 +481,7 @@ namespace eastl
 			inline void Move(Layout& dst, Layout& src) EA_NOEXCEPT       { eastl::swap(dst.raw, src.raw); }
 			inline void Swap(Layout& a, Layout& b) EA_NOEXCEPT           { eastl::swap(a.raw, b.raw); }
 
-			inline void ResetToSSO() EA_NOEXCEPT { memset(&raw, 0, sizeof(RawLayout)); }
+			inline void ResetToSSO() EA_NOEXCEPT { *SSOBeginPtr() = 0; SetSSOSize(0); }
 		};
 
 		eastl::compressed_pair<Layout, allocator_type> mPair;
@@ -935,6 +935,7 @@ namespace eastl
 	{
 		// Note that we do not call SizeInitialize here.
 		AllocateSelf(n);
+		internalLayout().SetSize(0);
 		*internalLayout().EndPtr() = 0;
 	}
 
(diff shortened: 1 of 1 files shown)


MaxEWinkler:
Hey, thanks for this.
Do you have any tests that exercise this segfault or do the current tests crash?
Do you have any assembly of the GCC9 code that segfaults. I'm curious about what it is doing. Is it the SetSSOSize(0) call?

MaxEWinkler:
I couldn't repro with the current tests on GCC 9.3.0, GCC 10.1.0, or Clang 10.
If you got a test case that shows the issue I would like to see.
I want to understand the core issue better and know whether or not we are doing anything that may be UB.

virt00l:
> I couldn't repro with the current tests on GCC 9.3.0, GCC 10.1.0, or Clang 10.
> If you got a test case that shows the issue I would like to see.
> I want to understand the core issue better and know whether or not we are doing anything that may be UB.

Hi. I did not tried to extract minimal repro for this issue (since I already buried too much time into this and my patch fixes it and potentially optimizes it as well).
AFAIK bug happens only on 9.X versions of GCC (i.e. 7,8 or 10 is not triggering it).
Actual code happens to look like 
bunch of calls to external function that accept one its argument with type "const string &foo = string()", i.e. something like
>bar(..., const string &foo = string());
with explicit c-strings (i.e. "foo" is implicitly constructed on stack from c string), e.g. something like
```
bar(...,"string");
bar(...,"string_length_24_or_more");
bar(...,"string_exact_length_23_");
```
last call produces broken SSO string without null-terminating null. 
Since string is SSO AND allocated on stack compiler seems to generate it's content not by load of string const (as one might expect), but by constructing it manually via combination of simd & integer stores, and seems to "forget" to write last null byte in case when string is exactly 23 characters long (so byte that supposed to be null-terminated seems to left from previous call (previous  stack-allocated string which was non-sso and therefore has 0x80 in place of null-terminator)))

I'm guessing that it somehow related to "dead store elimination" optimization, and default code (that e.g. memcpy everything in ctor) confused optimizer somehow. In any case this patch both fixes it and makes default ctor faster (as we write 2 bytes instead of 24)

virt00l:
Hm, I just realized, code that I described above is literally open-source and freely available for inspection:
https://github.com/GaijinEntertainment/daScript/blob/master/src/builtin/module_builtin_runtime_sort.cpp#L76
crash happens here, last argument of call to addExtern is exactly 23 character long - "builtin_sort_any_cblock", while previous is not ("builtin_sort_cblock<char *>")
> ```
> bar(...,"string");
> bar(...,"string_length_24_or_more");
> bar(...,"string_exact_length_23_");
> ```
>

MaxEWinkler:
Yep I agree. Thanks for the PR, its a definite good change :).

PR #382, opened 2020-07-16
Fix reverse adaptor when using on a rvalue range

lhamot:
Hello!
When trying the reverse adaptor on a rvalue container:

- There is this  warning on Visual 2015:

`warning C4239: nonstandard extension used : 'argument' : conversion from 'eastl::vector<int,eastl::allocator>' to 'eastl::vector<int,eastl::allocator> &'`

- And this error on Visual2017:

`error C2440: '<function-style-cast>': cannot convert from 'T' to 'eastl::reverse_wrapper<_Ty>'
        with
        [
            T=eastl::vector<int,eastl::allocator>
        ]
        and
        [
            _Ty=eastl::vector<int,eastl::allocator>
        ]
`

The member `Container& mContainer;` was wrong because on a lvalue *Container* it is still a reference (example: eastl::vector<int>&), and on a rvalue, it is just a copy (example: eastl::vector<int>) which is fine. So, no need to add a '&' to `Container mContainer;`.

### DIFF
--- a/include/EASTL/bonus/adaptors.h
+++ b/include/EASTL/bonus/adaptors.h
@@ -34,8 +34,9 @@ namespace eastl
 	template <typename Container>
 	struct reverse_wrapper
 	{
-		reverse_wrapper(Container& c) : mContainer(c) {}
-		Container& mContainer;
+		template <typename C>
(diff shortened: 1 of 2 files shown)


rparolin:
Thanks for the PR.

PR #386, opened 2020-08-25
Add eastl::span<*> to EASTL.natvis

Valakor:
This change just adds a natvis entry for the `span` class using `eastl::VectorBase` as a template. I've tested in VS2019 and it works as expected.

I've signed the CLA so I think that should be everything required?


rparolin:
Looks good to me. Thanks for the PR!

PR #392, opened 2020-10-13
rbtree::emplace_hint() compilation fix for variadic arg use

miherius:
This PR provides a fix for the `rbtree::emplace_hint()` method, when it called using multiply variadic template arguments. The problem is in `DoInsertValueHint()` method, which can't accept the forwarded parameter pack, expecting an already constructed `pair` only. A small example:
```cpp
eastl::map<int, char> m;
m.emplace(1, 'a'); // ok, arguments are forwarded to the `pair` constructor
m.emplace_hint(m.end(), 2, 'b'); // compilation fail, arguments can't be forwarded
m.emplace_hint(m.end(), eastl::make_pair(2, 'b')); // ok, passing a constructed `pair`
```
This PR makes the interface of the `DoInsertValueHint()` method (internal helper for "hint" versions of insert methods) similar to the `DoInsertValue()` method (another helper for non-"hint" insert methods), so it can be used with variadic template arguments now.

Additional fixes:
* `rbtree::try_emplace()` move rvalue-ref `key` instead of copy;
* `rbtree::try_emplace()` delegates `value_type` construction to the internal logic;
* "hint"-version of non-unique `hashtable::insert()` can be complied now (was needed for new tests).

Map and multimap tests have been extended to cover more use-cases of these methods (changes in `TestMapCpp11()`, `TestMultimapCpp11()` functions).

There are no changes in the public interface of the mentioned classes.

### DIFF
--- a/include/EASTL/internal/hashtable.h
+++ b/include/EASTL/internal/hashtable.h
@@ -2812,7 +2812,7 @@ namespace eastl
 	{
 		// We ignore the first argument (hint iterator). It's not likely to be useful for hashtable containers.
 		insert_return_type result = DoInsertValue(has_unique_keys_type(), value);
-		return result.first; // Note by Paul Pedriana while perusing this code: This code will fail to compile when bU is false (i.e. for multiset, multimap).
+		return DoGetResultIterator(has_unique_keys_type(), result);
 	}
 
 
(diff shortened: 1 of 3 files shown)


rparolin:
Thanks for the PR!  Looks go to me.

PR #395, opened 2020-11-05
[EASTL 3.17.02]

MaxEWinkler:
eastl::atomic<T>

- fix all the spelling mistakes in the doc
- Added support for non-trivially default constructible types
- Cleaned up comments and impl
- improved 128-bit load code gen
- fixed type pun to support non-trivially default constructible types
- ensure msvc instrinics do not emit prefetch instructions

EASTL:  to_array implementation

EASTL: fix for rbtree input iterator ctor moving elements from the source container


dAu6jARL:
Please update EASTL_VERSION in internal/config.h

PR #400, opened 2020-11-20
Fixed warnings in function_detail.h on MSVC at warning level 4.

bleachkitty:
Fixed warning C4100: 'functor': unreferenced formal parameter and warning C4100: '<args_0>': unreferenced formal parameter on MSVC with warning level 4.


rparolin:
Thanks for the PR!

PR #405, opened 2020-11-27
Using of qualified eastl move() and forward() functions.

miherius:
This PR is similar to #333 in motive. In case of `std` types used as elements of some `eastl` containers, argument-dependent lookup result in `ambiguous call` errors for `move()` and `forward()` calls.

Small example: line `auto [str, flag] = eastl::pair<std::string, bool>("", false);` can't be compiled.


rparolin:
Looks good.  Thanks for the PR!

PR #406, opened 2020-12-03
Fix get<T>(variant) when variant is a rvalue

lhamot:
Hello!
The function `eastl::get<T>(variant&&)` can't work because the `get<I>(v)` take a lvalue (*v* has a name so it is a lvalue) and so return a lvalue, which can't be implicitly converted to rvalue.
`eastl::move` is used to force *v* to be a rvalue.


rparolin:
Looks good to me.  Thanks for the patch!

PR #407, opened 2020-12-11
Using of qualified eastl functions.

miherius:
This PR makes calls to some eastl functions qualified. This helps to avoid ambiguity in a function calls caused by argument-dependent lookup.
Full list of functions affected by this PR:
```
clamp
max
find
next_permutation
is_sorted_until
rbegin
rend
addressof
destroy_at
move
forward
forward_as_tuple
```


rparolin:
LGTM!  Thanks for the pull request!

PR #409, opened 2020-12-23
Removed extra parentheses on eastl::move's return statement

Razakhel:
This silences a specific clang-tidy check when using the pass-by-value + move idiom

```cpp
class Foo
{
public:
    Foo(eastl::string str)
        : m_str{ eastl::move(str) }
    {}

private:
    eastl::string m_str;
};
```
```
warning: the parameter 'str' is copied for each invocation but only used as a const reference; consider making it a const reference [performance-unnecessary-value-param]
        Foo(eastl::string str)
                          ^
            const        &
```

With the change introduced with this PR, this warning disappears.


MaxEWinkler:
Looks good. Thanks for great find :).

PR #411, opened 2021-01-04
Added structured binding support for eastl::array

Razakhel:
This is based off what has already been done for eastl::pair.
The changes are tested, checking that overriding a bound variable does modify the original array's element, and that the constness is preserved.


james-moran-ea:
Hi there,
Sorry for the lack of movement on this.  This doesn't raise any red flags to me, thanks for the submission!
James

PR #414, opened 2021-02-17
Ability to default-construct tuple containing an empty type.

miherius:
Currently, attempt to default-construct tuple with an empty type results in a compiler error. Reason is in specialization of internal helper class `TupleLeaf` which implicitly deletes default constructor in case of `is_empty_v<ValueType>`.


miherius:
Problem fixed in version 3.21.12

Issue #387, opened 2020-08-29
Eastl in windows kernel driver 


AlexCasual:
Can I use eastl in the Windows kernel driver without exceptions in containers (vector, list, map) and strings (unicode)?
Help me start using eastl in the kernel)

DBJDBJ:
### [https://github.com/DBJDBJ/EASTL2010](https://github.com/DBJDBJ/EASTL2010)

PR #415, opened 2021-02-22
Add assertion failure variant for win kernel

mtnpke:
Like some others (#118, #387), I am using EASTL in Windows kernel mode. Some of the default source files such as `assert.cpp` do not compile/work in this context because they require user-mode Windows functions. This PR removes this dependency from `assert.cpp` and provides a default variant for assertion failure in kernel mode.

This needs electronicarts/EABase#4 as prerequisite.

### DIFF
--- a/source/assert.cpp
+++ b/source/assert.cpp
@@ -7,7 +7,9 @@
 #include <EASTL/string.h>
 #include <EABase/eabase.h>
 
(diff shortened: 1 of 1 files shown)


james-moran-ea:
I don't see any red flags with this, I've merged this and the EABase changes. Thanks for the submission!

PR #422, opened 2021-04-18
Fix undefined behavior in memory.h

eugeneko:
This is a fix for crashes in optimized builds by GCC.

Literally the same bug and the same fix as in #280, just in two more functions


liam-mitchell:
Looks good to me, thanks for the PR.

PR #424, opened 2021-04-22
Added ring_buffer::push_back/push_front accepting a rvalue reference

matt77hias:
Not sure whether the `ring_buffer::push_back` and `ring_buffer::push_front` in general should use (copy and move) assignment instead of in place (copy and move) construction in order to be more consistent with other containers.

### DIFF
--- a/include/EASTL/bonus/ring_buffer.h
+++ b/include/EASTL/bonus/ring_buffer.h
@@ -282,9 +282,11 @@ namespace eastl
 		const_reference back() const;
 
 		void            push_back(const value_type& value);
+		void            push_back(value_type&& value);
 		reference       push_back();
 
 		void            push_front(const value_type& value);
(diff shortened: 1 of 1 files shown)


include/EASTL/bonus/ring_buffer.h
```diff
@@ -1192,6 +1194,27 @@ namespace eastl
 	}
 
 
+	/// A push_back operation on a ring buffer assigns the new value to end.  
+	/// If there is no more space in the buffer, this will result in begin
+	/// being overwritten and the begin position being moved foward one position.
```
stingoh, line 1199: Minor typo: "foward" should be "forward"

stingoh: (although this typo is present everywhere else)

PR #436, opened 2021-08-05
Update FAQ.md

ezhikus:
Fixed double mention of push_back(void) in description of additional functions of the list container (push_front should be mentioned)


liam-mitchell:
Good catch, thanks for the PR.

PR #438, opened 2021-08-06
Fixes typo

jgressmann:


liam-mitchell:
Good catch, thanks for the PR.

PR #442, opened 2021-09-14
Replaced some generic calls to operator< by eastl::less

Razakhel:
Generalized the use of `eastl::less` in several sorting-related functions. Its usage should be preferred in algorithms, to allow users specializing it for their classes instead of having to define a public operator<.
I've only fixed errors occuring in my specific case; there may be other occurrences elsewhere.

This is also applicable to `eastl::equal_to` (operator==), `eastl::not_equal_to` (operator!=), `eastl::greater` (operator>), `eastl::greater_equal` (operator>=) and `eastl::less_equal` (operator<=). Unless specialized, they all fallback on their respective operator.

### DIFF
--- a/include/EASTL/algorithm.h
+++ b/include/EASTL/algorithm.h
@@ -806,18 +806,18 @@ namespace eastl
 	template <typename T>
 	inline T&& median_impl(T&& a, T&& b, T&& c)
 	{
-		if(a < b)
+		if(eastl::less<T>()(a, b))
 		{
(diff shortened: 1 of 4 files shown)


james-moran-ea:
Hi there,
We just pushed a new CI workflow that uses github actions since our existing travis CI hasn't been running on new PRs. Could you sync down the latest changes and update this PR? Once those checks succeed, we can move forward with this :)
Thanks!

PR #443, opened 2021-10-13
Honour EA_NO_HAVE_CPP11_TUPLES before trying to include <tuple>

zhmu:
EA_COMPILER_NO_STRUCTURED_BINDING does not properly check whether <tuple> can be included.

This change strengthens the condition to prevent build errors on C++17 when <tuple> is not available.


james-moran-ea:
Hi there,
Apologies for the delay in getting back to you, we appreciate the submission. We just pushed a new CI workflow that uses github actions since our existing travis CI hasn't been running on new PRs. Could you sync down the latest changes and update this PR so that the checks run? Once those changes are made and the checks succeed, we can move forward with this :)
Thanks again!

PR #444, opened 2021-10-13
Do not use file(GLOB ...) to collect source files

zhmu:
This is specifically recommended against in the CMake documentation [1]:

"We do not recommend using GLOB to collect a list of source files from your
source tree. [...]"

[1] https://cmake.org/cmake/help/v3.15/command/file.html#filesystem


james-moran-ea:
Hi there,
We just pushed a new CI workflow that uses github actions since our existing travis CI hasn't been running on new PRs. Could you sync down the latest changes and update this PR? Once those checks succeed, we can move forward with this :)
Thanks!

Issue #441, opened 2021-09-04
Typo in eastl::pair tests


eyalroz:
On this line:
https://github.com/electronicarts/EASTL/blob/95b281ee345372655f189ea5baa50adb3a61832a/test/source/TestUtility.cpp#L111

the comparison should be of `p3.second`, not `p2.second`.

Just noticed that as I was lifting your pair class for my [CUDA Kernel Author's Toolkit](https://github.com/eyalroz/cuda-kat/).

PR #453, opened 2021-12-09
Fix typo in unit test

GrooveStomp:
Use string compare on the object created for the test; not the previously created object.

https://github.com/electronicarts/EASTL/issues/441


james-moran-ea:
Hi there,
We just pushed a new CI workflow that uses github actions since our existing travis CI hasn't been running on new PRs. Could you sync down the latest changes and update this PR. Once those checks succeed, we can move forward with this :)
Thanks!

grojo-ea:
This change looks safe enough to merge without the CI running on it. Thanks for the fix.

PR #459, opened 2022-02-27
atomic macros: replace static_assert(false) condition by a dependent type always false (fix VS 2022 compilation error)

Wawha:
With VS 2022, `static_assert(false)` will always failed, event inside template code never instantiate.
A dependent type need to be use.
See https://docs.microsoft.com/en-us/cpp/overview/cpp-conformance-improvements?view=msvc-170#error-on-a-non-dependent-static_assert

If there is a better name for `eastl_atomic_always_false` tell me

### DIFF
--- a/include/EASTL/internal/atomic/atomic_macros/atomic_macros_base.h
+++ b/include/EASTL/internal/atomic/atomic_macros/atomic_macros_base.h
@@ -10,6 +10,8 @@
 	#pragma once
 #endif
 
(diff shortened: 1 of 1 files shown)


grojo-ea:
Hello, 

Thanks a lot for the PR, we had an equivalent internal fix for this issue that just got merged in #477 so I'm closing this PR for now given that this overlaps with our fix.

PR #460, opened 2022-02-28
map::at(): exception must be throw (or assertion) if element, smaller than the first one, is not present

Wawha:
map::at() do throw/generate error if at is call on value smaller than the first element.
```c++
		eastl::map<int, int> m;
		m.emplace(2, 85);
		auto v = m.at(0); // Must throw
```

### DIFF
--- a/include/EASTL/map.h
+++ b/include/EASTL/map.h
@@ -445,7 +445,7 @@ namespace eastl
 	{
 		iterator itLower(lower_bound(key)); // itLower->first is >= key.
 
-		if(itLower == end())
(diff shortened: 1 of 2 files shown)


Wawha:
Note that compilation error will be solved when this PR will be merged: https://github.com/electronicarts/EASTL/pull/459

grojo-ea:
Hello,

Thanks a lot for the PR, we had an equivalent internal fix for this issue that just got merged in #477 so I'm closing this PR for now given that this overlaps with our fix.

PR #469, opened 2022-05-23
Fixing issue with list ctor that differs from stl (wasn't documented)

RisingLiberty:
(https://en.cppreference.com/w/cpp/container/list/list)
constructor 4) dictates: "Constructs the container with count default-inserted instances of T. No copies are made."
However, EASTL created a default constructed object which would then be copied into the nodes.
No default object is constructed now and the parameter-less overload of DoCreateNode is now used to create the nodes.

### DIFF
--- a/test/source/TestList.cpp
+++ b/test/source/TestList.cpp
@@ -54,6 +54,12 @@ int TestList()
 
 		VERIFY(eastl::all_of(l.begin(), l.end(), [](int e)
 		                     { return e == 0; }));
+
+		TestObject::Reset();
+		eastl::list<TestObject> l2(test_size);
+		VERIFY(TestObject::sTOCtorCount == test_size);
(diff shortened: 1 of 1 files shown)


grojo-ea:
Hello, 

Thanks for the PR, what's here  looks good to me but it looks like we don't have a [CLA](https://github.com/electronicarts/EASTL/blob/master/CONTRIBUTING.md) for you on file yet. Could you please fill out [this](https://electronicarts.na1.echosign.com/public/esignWidget?wid=CBFCIBAA3AAABLblqZhByHRvZqmltGtliuExmuV-WNzlaJGPhbSRg2ufuPsM3P0QmILZjLpkGslg24-UJtek*) form?

Also, it'd be good to have a test to `TestList.cpp`  verify that the "No copies are made" part of the contract is satisfied. You should be able to verify this via `TestObject::sTOCopyCtorCount`.

PR #470, opened 2022-05-26
Removed (reduntant) template id for destructor.

heliobatimarqui:
The EASTL doesn't compile when using C++ 20 in GCC 11.2.1, because of line 1724. This modification makes it compile when using -std=c++20.


grojo-ea:
Hello,

Thanks for the PR. We've recently changed some of the code around `eastl::is_destructible` which in particular should have fixed the issue this PR addresses. The fix is in #477.

PR #474, opened 2022-07-05
Avoid using `_FInf` etc. in `numeric_limits.h` after MSVC's STL removes them

StephanTLavavej:
I work on the Visual C++ team, where we regularly build many open-source projects, including EASTL, with development builds of the MSVC compiler and libraries, in order to prevent regressions that would affect projects. This also allows us to provide advance notice of breaking changes, such as this one:

We've merged https://github.com/microsoft/STL/pull/2828, which is removing the declarations of the `_FInf` etc. variables from our headers. This change will ship in VS 2022 17.4 Preview 2, and it breaks EASTL because MSVC's STL still defines `_CPPLIB_VER` (identifying ourselves as Dinkumware-derived sources).

This PR uses our supported macro [`_MSVC_STL_UPDATE`](https://github.com/microsoft/STL/wiki/Macro-_MSVC_STL_UPDATE) for identifying the updated versions of MSVC's STL, and instead uses the compiler's `__builtin_huge_valf()` etc. intrinsics, plus `FLT_TRUE_MIN` etc. from the UCRT.

These intrinsics and macros have been available for quite some time. Note that they will improve codegen, as they produce constants that are known to the compiler. In contrast, the old internal variables are separately compiled (and modifiable!), which inhibits optimizations like constant propagation.

### DIFF
--- a/include/EASTL/numeric_limits.h
+++ b/include/EASTL/numeric_limits.h
@@ -1435,6 +1435,19 @@ namespace eastl
 			static value_type round_error() 
 				{ return 0.5f; }
 
+			#if defined(_MSVC_STL_UPDATE) && _MSVC_STL_UPDATE >= 202206L // If using a recent version of MSVC's STL...
+			static value_type infinity()
+				{ return __builtin_huge_valf(); }
+
+			static value_type quiet_NaN()
+				{ return __builtin_nanf("0"); }
+
+			static value_type signaling_NaN()
+				{ return __builtin_nansf("1"); }
+
+			static value_type denorm_min()
+				{ return FLT_TRUE_MIN; }
+			#else
 			static value_type infinity() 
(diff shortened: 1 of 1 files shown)


grojo-ea:
Hi @StephanTLavavej, thank you very much for brining this to our attention and for the PR with the fix. 

We'd like to take this PR but it looks like we don't have a Contributor License Agreement for you on file yet. Do you mind filling out [this](https://electronicarts.na1.echosign.com/public/esignWidget?wid=CBFCIBAA3AAABLblqZhByHRvZqmltGtliuExmuV-WNzlaJGPhbSRg2ufuPsM3P0QmILZjLpkGslg24-UJtek*) form so we can merge this and any future PRs that you author? 

Thanks again.

StephanTLavavej:
@grojo-ea Done, thanks!

PR #476, opened 2022-07-17
Use invoke_result instead of result_of in reference_wrapper.

eugeneko:
It fixes compilation when `eastl::ref(object)` is used as callable object in `eastl::is_invokable`.

Here is code example (unless I messed up because I am writing this code from memory)
```
eastl::function<void()> foo;
constexpr bool v1 = eastl::is_invocable_r_v<void, decltype(eastl::ref(foo))>; // ok
constexpr bool v2 = eastl::is_invocable_r_v<void, decltype(eastl::ref(foo)), int>; // compilation error
```

CppReference also uses `invoke_result` in example implementation.
https://en.cppreference.com/w/cpp/utility/functional/reference_wrapper

### DIFF
--- a/include/EASTL/internal/functional_base.h
+++ b/include/EASTL/internal/functional_base.h
@@ -118,7 +118,8 @@ namespace eastl
 
 	template <typename R, typename F, typename... Args>
 	struct is_invocable_r_impl<R, F, void_t<typename invoke_result<F, Args...>::type>, Args...>
-		: public is_convertible<typename invoke_result<F, Args...>::type, R> {};
+		: public disjunction<is_convertible<typename invoke_result<F, Args...>::type, R>,
+							 is_same<typename remove_cv<R>::type, void>> {};
 
 	template <typename R, typename F, typename... Args>
 	struct is_invocable_r : public is_invocable_r_impl<R, F, void, Args...> {};
@@ -232,7 +233,7 @@ namespace eastl
 		T& get() const EA_NOEXCEPT;
 
 		template <typename... ArgTypes>
-		typename eastl::result_of<T&(ArgTypes&&...)>::type operator() (ArgTypes&&...) const;
+		typename eastl::invoke_result<T&, ArgTypes...>::type operator() (ArgTypes&&...) const;
 
 	private:
 		T* val;
@@ -269,7 +270,7 @@ namespace eastl
 
 	template <typename T>
 	template <typename... ArgTypes>
-	typename eastl::result_of<T&(ArgTypes&&...)>::type reference_wrapper<T>::operator() (ArgTypes&&... args) const
+	typename eastl::invoke_result<T&, ArgTypes...>::type reference_wrapper<T>::operator() (ArgTypes&&... args) const
 	{
 		return eastl::invoke(*val, eastl::forward<ArgTypes>(args)...);
 	}
--- a/test/source/TestFunctional.cpp
+++ b/test/source/TestFunctional.cpp
@@ -1491,5 +1491,39 @@ struct TestInvokeResult
 };
 
 template struct eastl::invoke_result<decltype(&TestInvokeResult::f), TestInvokeResult, void>;
+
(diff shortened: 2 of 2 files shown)


grojo-ea:
Hello,

Thanks for the PR, this change looks good to me but could you please add some tests that verify the issue this is fixing. Perhaps put them in `TestFunctional.cpp`. It looks like we don't currently have any test for calling `operator()` on callables that are wrapped by `reference_wrapper` so adding some there would be good.

eugeneko:
@grojo-ea Done

test/source/TestFunctional.cpp
```diff
@@ -1491,5 +1491,39 @@ struct TestInvokeResult
 };
 
 template struct eastl::invoke_result<decltype(&TestInvokeResult::f), TestInvokeResult, void>;
+
 static_assert(!eastl::is_invocable<decltype(&TestInvokeResult::f), TestInvokeResult, void>::value, "incorrect value for is_invocable");
+static_assert(!eastl::is_invocable<decltype(&TestInvokeResult::f), TestInvokeResult, int, int>::value, "incorrect value for is_invocable");
 static_assert(eastl::is_invocable<decltype(&TestInvokeResult::f), TestInvokeResult, int>::value, "incorrect value for is_invocable");
+
+static_assert(!eastl::is_invocable_r<int, decltype(&TestInvokeResult::f), TestInvokeResult, void>::value, "incorrect value for is_invocable_r");
+static_assert(!eastl::is_invocable_r<void, decltype(&TestInvokeResult::f), TestInvokeResult, int>::value, "incorrect value for is_invocable_r");
```
grojo-ea, line 1500: I believe this goes against the spec and it's a bug in `eastl::is_invocable_r` (a different bug from the one this PR is trying to fix). From the docs on [std::is_invocable_r](https://en.cppreference.com/w/cpp/types/is_invocable)

>  ... If R is cv void, the result can be any type...

I've verified this by plugging these tests into compiler explorer [here](https://godbolt.org/z/sxMcGjbM4). 

I think the fix is to add a disjunction in `is_invocable_r_impl`, something like this:
```
	template <typename R, typename F, typename... Args>
	struct is_invocable_r_impl<R, F, void_t<typename invoke_result<F, Args...>::type>, Args...>
	    : public disjunction<is_convertible<typename invoke_result<F, Args...>::type, R>,
	                         is_same<typename remove_cv<R>::type, void>> {};
```
Whether you want to put that fix in as part of this PR or not is up to you. I think it'd be good to iron that out, but if you decide not to put the fix in then I'd ask to please correct the expected behavior in this test (i.e. remove the `!` from this static assert, and the similar ones below) and comment it out with a clarifying comment saying that the test is disabled due to a bug in `eastl::is_invocable_r` not conforming to the spec when `void` is passed as a return value.

include/EASTL/internal/functional_base.h
```diff
@@ -232,7 +232,7 @@ namespace eastl
 		T& get() const EA_NOEXCEPT;
 
 		template <typename... ArgTypes>
-		typename eastl::result_of<T&(ArgTypes&&...)>::type operator() (ArgTypes&&...) const;
+		typename eastl::invoke_result<T&, ArgTypes&&...>::type operator() (ArgTypes&&...) const;
```
grojo-ea, line 235: I'm having trouble coming up with a case where it makes a difference, perhaps it doesn't, but I don't think the `&&` in `<T&, ArgTypes&&...>` is necessary. If we can get away with a bit less noise I recommend removing it here and below where it's implemented.
```suggestion
		typename eastl::invoke_result<T&, ArgTypes...>::type operator() (ArgTypes&&...) const;
```

grojo-ea (changes requested):
Thanks for adding the tests. I think you've actually uncovered an issue with `eastl::in_invocable_r`. Up to you if you want to fix that as part of this PR or not but I'd prefer not to have enabled tests that explicitly assert behavior that's incorrect according to the specification unless we are doing it intentionally and have a good reason to diverge (which I don't believe we have in this case).

eugeneko:
Done (2)

PR #477, opened 2022-07-19
Release candidate for EASTL version 3.19.03

grojo-ea:
**Feature additions:**

- Added eastl::to_address helper function in <memory>.
- Added midpoint and lerp algorithms.
- Added integral power of 2 operations.
- Added safe integral comparisons.
- Added three-way-comparison operator for all eastl types.
- Implemented is_detected idiom into the header <EASTL/type_traits.h> and the namespace eastl:
    https://en.cppreference.com/w/cpp/experimental/is_detected 
- Added a bonus helper class "overloaded" to simplify variant visits.

**Improvements:**

- Use integer sequence builtins from various compilers to improve make_integer_sequence.
- Made eastl::iterator_traits SFINAE friendly as required by N3844.
- Remove dependency on Windows.h in atomic
- Changed the default natvis visualization of hashtable to be more user friendly.
  - The previous hashtable natvis view is accessible by adding ",view(detailed)" in the watch window
- Tuples can now contain l-value references to incomplete types.
- Added VS natvis support for eastl::tuple.

**Bug-fixes:**

- Fixed `map::at` so it properly throws when the key is not in the map.
- Fixing `reverse_iterator` unwrapping:
  - `is_iterator_wrapper` now relies on the presence of an `unwrap()` member function which returns the unwrapped iterator (instead of the `wrapped_iterator_type` typedef).
  - `reverse_iterator<Iter>` is considered a wrapped iterator if and only if `Iter` is a wrapped iterator (the previous behaviour was that `reverse_iterator<Iter>` was always considered wrapped).
  - Unwrapping a `reverse_iterator<Iter>` now returns a `reverse_iterator<T>` where `T` is the type resulting from unwrapping `Iter`. This fixes various bugs related to assigning from reverse iterators, for which there's now new unit tests.
- Fixed is_destructible, is_trivially_destructible and is_nothrow_destructible:
  - r-value and l-value references are considered as destructibles.
  - abstract classes are considered as destructibles.
  - MSVC relies on the builtins __is_destructible, __is_trivially_destructible and __is_nothrow_destructible when possible.
- Fixed a bug that made VS natvis work incorrectly for some small strings.
- Fixed a bug that could cause radix sort to sort incorrectly on some input
- Fix atomic macros so they compile with /Zc:static_assert enabled in MSVC 17.10.
- Fixed try `try_emplace` for `hash_map` and `map` so it does not create a node when the key already exists.
- Removed `try_emplace` from `hash_(multi)set` and `(multi)set`, it was broken (i.e did not compile) and shouldn't be part of the API.
- Fixed some eastl instead of EASTL_ITC_NS occurences.
- Fixed compilation error with certain operator overloads in chrono.h.
- Fixed warning in chrono.h on MSVC with high warning levels.


SamVanheer:
Is this scheduled for release or is there still more work to be done? Would love to get this on vcpkg so C++20 builds with GCC 11 work again.

PR #481, opened 2022-08-22
Add non-explicit default ctor for hash_map and hash_set.

eugeneko:
It's annoying that default ctors of map and set are explicit, because the following code doesn't compile:

```
struct Test
{
    eastl::unordered_map<int, int> m;
};

Test t = {};
```
and I **must** do `m{};`

### DIFF
--- a/include/EASTL/hash_map.h
+++ b/include/EASTL/hash_map.h
@@ -125,8 +125,19 @@ namespace eastl
 		///
 		/// Default constructor.
 		///
-		explicit hash_map(const allocator_type& allocator = EASTL_HASH_MAP_DEFAULT_ALLOCATOR)
-			: base_type(0, Hash(), mod_range_hashing(), default_ranged_hash(), 
+		hash_map()
+			: this_type(EASTL_HASH_MAP_DEFAULT_ALLOCATOR)
+		{
+			// Empty
+		}
+
+
+		/// hash_map
+		///
+		/// Constructor which creates an empty container with allocator.
+		///
+		explicit hash_map(const allocator_type& allocator)
+			: base_type(0, Hash(), mod_range_hashing(), default_ranged_hash(),
 						Predicate(), eastl::use_first<eastl::pair<const Key, T> >(), allocator)
 		{
 			// Empty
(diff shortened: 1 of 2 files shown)


include/EASTL/hash_map.h
```diff
@@ -125,8 +125,20 @@ namespace eastl
 		///
 		/// Default constructor.
 		///
-		explicit hash_map(const allocator_type& allocator = EASTL_HASH_MAP_DEFAULT_ALLOCATOR)
-			: base_type(0, Hash(), mod_range_hashing(), default_ranged_hash(), 
+		hash_map()
+			: base_type(0, Hash(), mod_range_hashing(), default_ranged_hash(),
+						Predicate(), eastl::use_first<eastl::pair<const Key, T> >(), EASTL_HASH_MAP_DEFAULT_ALLOCATOR)
```
grojo-ea, line 130: Probably better to not duplicate code and do `: has_map(EASTL_HASH_MAP_DEFAULT_ALLOCATOR)` here?

include/EASTL/hash_map.h
```diff
@@ -125,8 +125,20 @@ namespace eastl
...
+		{
+			// Empty
+		}
+
+
+		/// hash_map
+		///
+		/// Default constructor.
```
grojo-ea, line 138: This is no longer the default constructor.

include/EASTL/hash_set.h
```diff
@@ -117,8 +117,19 @@ namespace eastl
 		/// hash_set
 		///
 		/// Default constructor.
-		/// 
-		explicit hash_set(const allocator_type& allocator = EASTL_HASH_SET_DEFAULT_ALLOCATOR)
+		///
+		hash_set()
+			: base_type(0, Hash(), mod_range_hashing(), default_ranged_hash(), Predicate(), eastl::use_self<Value>(), EASTL_HASH_SET_DEFAULT_ALLOCATOR)
```
grojo-ea, line 122: Same as above, reduce code duplication by forwarding to the constructor taking an allocator.

include/EASTL/hash_set.h
```diff
@@ -117,8 +117,19 @@ namespace eastl
...
+		{
+			// Empty
+		}
+
+
+		/// hash_set
+		///
+		/// Default constructor.
```
grojo-ea, line 130: Same as above, this is not the default constructor anymore.

grojo-ea:
Thanks for the fixes

PR #423, opened 2021-04-19
Add lost install commands

shossjer:
Not having an install target makes it difficult to use EASTL in an automatic build setup. These two lines of code were removed from EASTL in 376bcdc93502208b3e61b88c911fad0f9c9e3f60. This is not by any means new code. Since this, then, clearly is the work of others (and not my own) I cannot in good confidence sign the EA CLA.

This should be merged after https://github.com/electronicarts/EABase/pull/5.

PR #486, opened 2022-09-09
Re-add install commands

SamVanheer:
This adds back the install commands that were removed in an earlier commit.

This is the same as #423 but does not include any other changes and is based on the current head commit.


grojo-ea:
Hello, 
Thanks for the PR.

PR #487, opened 2022-09-09
Add EASTL.natvis to EASTL when using MSVC and install to doc directory

SamVanheer:
This adds `EASTL.natvis` to the EASTL target when using MSVC and installs the file to the `doc` directory, enabling users to use the file without needing to reference the Git repository or manually copy the file.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -23,6 +23,15 @@ include(CommonCppFlags)
 file(GLOB EASTL_SOURCES "source/*.cpp")
 add_library(EASTL ${EASTL_SOURCES})
 
+if (MSVC)
+    set(EASTL_NATVIS_DIR "doc")
+    set(EASTL_NATVIS_FILE "${EASTL_NATVIS_DIR}/EASTL.natvis")
+    target_sources(EASTL INTERFACE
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -22,6 +22,15 @@ include(CommonCppFlags)
 file(GLOB EASTL_SOURCES "source/*.cpp")
 add_library(EASTL ${EASTL_SOURCES})
 
+if (MSVC)
+	set(EASTL_NATVIS_DIR "doc")
```
grojo-ea, line 26: Could you fix this indentation?

grojo-ea:
Thanks for the PR. This looks good to me but it now needs a rebase.

CMakeLists.txt
SamVanheer: I've normalized the indentation to use 4 spaces for all new additions, but it looks like indentation in existing files is also inconsistent. Let me know it needs changing.

SamVanheer:
I've rebased it and applied the requested changes.

CMakeLists.txt
grojo-ea: Good point, we should probably fix that indentation at some point, but that's separate from this PR. Thanks for fixing it up for the added lines.

PR #488, opened 2022-10-12
remake vector_set from inheritance from underlying container to aggre…

AntonYudintsev:
…gation

this makes it safer, as it disallows some of changes that could violate sorted contract it also saves at least sizeof(void*) on most common comparators (due to empty struct optimization in compressed_pair)

con: it explicitly requires container class to follow vector API (almost all functions are exposed in vector_set)

to consider: remove non const front(), back(), at(), [] access. Those are not safe either, left them only for compatibility


grojo-ea:
Hello,

We actually have an internal change in progress (unreleased as of this writing) to reduce the size of `vector_(multi)set` and `vector_(multi)map` that I think will conflict with this change so I'm going to hold off on the PR for now.

grojo-ea:
Hello again,

Version 3.21.12 (https://github.com/electronicarts/EASTL/pull/510) has the changes we had in progress that I mentioned above. Let me know if you want to try to rebase on the new version to try to get some of the API/inheritance changes reviewed or if we should close this PR instead.

PR #490, opened 2022-10-13
bitvector: remove explicit copy constructor and copy assignment

AntonYudintsev:
not only it is same as default implementation, but also disallows move assignment (violates rule-of-5)

since, there is no sense in declaring copy/assignment, we just remove it (use rule of zero) https://en.cppreference.com/w/cpp/language/rule_of_three


grojo-ea:
Good find, and thanks for the PR!

Issue #454, opened 2021-12-17
Small fixed_strings do not overflow correctly


youdrovemetoit:
The header comment for fixed string currently says 

	/// A fixed_string with bEnableOverflow == true is identical to a regular
	/// string in terms of its behavior.

But overflow behaviour appears to be broken when the fixed string size is small. For example this test case 

    eastl::fixed_string<char, 8> test;
    //eastl::string test;
    
    for (int i = 0; i < 12; ++i)
    {
        test.append(1, i % 10 == 0 ? 'a' : 'b');
        std::cout << test.c_str() << std::endl;
    }

produces

    a
    ab
    abb
    abbb
    abbbb
    abbbbb
    abbbbbb
    abbbbbb
    abbbbbb
    abbbbbb
    abbbbbb
    abbbbbb

This does not happen for string. The function is still appending characters, but it hasn't moved the null so the string ends early when output.
This appears to happen because fixed strings are created on the heap by default, regardless of size.

	internalLayout().SetHeapBeginPtr(mArray);
	internalLayout().SetHeapCapacity(nodeCount - 1);
	internalLayout().SetHeapSize(0);

If the user then appends another element, causing it to overflow, the string function

    inline void basic_string<T, Allocator>::set_capacity(size_type n)

is ultimately called for the new capacity. The new capacity is below the SSO capacity, and it runs the code following the comment

    // heap->sso
    // A heap based layout wants to reduce its size to within sso capacity
    // An sso layout wanting to reduce its capacity will not get in here

The problem is this assumption is incorrect for this use case, the size is not reducing it's growing, and the following code does not work correctly. Worse, this code

    CharStringUninitializedCopy(pOldBegin, pOldBegin + n, internalLayout().SSOBeginPtr());

seems to read beyond the end of the old heap size when copying into the SSO buffer, as n is the new larger data size.

I'm not sure what the correct solution is in this case, as I don't know the reason for fixed_string using the heap is for small allocations. I can bodge a min into the set_capacity function

    const size_type nMinSize = min(nOldCap, n);
    
    CharStringUninitializedCopy(pOldBegin, pOldBegin + nMinSize, internalLayout().SSOBeginPtr());
    internalLayout().SetSSOSize(nMinSize);

[...]

PR #495, opened 2022-12-03
Fix fixed_string resize within SSO capacity

W4RH4WK:
As pointed out in #454, when fixed_string gets resized but capacity stays within SSO capacity the null-terminator gets copied over as well. The corresponding code assumes heap capacity is always greater than SSO capacity. fixed_string's internal buffer is handled as heap inside the string class and may, in fact, be <= SSO capacity.

This commit fixes the described issue and adds a unit test for this specific case.

### DIFF
--- a/include/EASTL/string.h
+++ b/include/EASTL/string.h
@@ -1415,8 +1415,14 @@ namespace eastl
 					pointer pOldBegin = internalLayout().BeginPtr();
 					const size_type nOldCap = internalLayout().GetHeapCapacity();
 
-					CharStringUninitializedCopy(pOldBegin, pOldBegin + n, internalLayout().SSOBeginPtr());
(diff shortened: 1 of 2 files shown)


grojo-ea:
Hello, 

Thanks for looking at this issue and for the PR. I'm going to have to think a bit about this to see if this is the way we want to fix the issue. 

In the meantime, I've noticed we don't have a CLA from you yet, could you please go [here](https://electronicarts.na1.echosign.com/public/esignWidget?wid=CBFCIBAA3AAABLblqZhByHRvZqmltGtliuExmuV-WNzlaJGPhbSRg2ufuPsM3P0QmILZjLpkGslg24-UJtek*) and fill out the form? We'll need it if we decide this is the way to go to solve the issue.

PR #503, opened 2023-01-30
Add MSVC support to initializer_list

mtnpke:
The C++ standard does not define how the implementation constructs initializer_list instances from the actual lists. Many compilers seem to use the pointer and size method that EASTL implements, but not MSVC. It uses two pointers (the other obvious way to do it). This PR implements the corresponding constructor to make intializer_list compatible with MSVC.

### DIFF
--- a/include/EASTL/initializer_list.h
+++ b/include/EASTL/initializer_list.h
@@ -56,6 +56,14 @@
 			initializer_list() EA_NOEXCEPT  // EA_NOEXCEPT requires a recent version of EABase.  
 			  : mpArray(NULL), mArraySize(0) { }
 
+#if defined(EA_COMPILER_MSVC)
+			// MSVC generates constructor calls with two pointers instead of one pointer + size. The constructor is
+			// public.
+			// See: https://docs.microsoft.com/en-us/cpp/standard-library/initializer-list-class#initializer_list
+			initializer_list(const_iterator pFirst, const_iterator pLast) EA_NOEXCEPT
+			  : mpArray(pFirst), mArraySize(pLast - pFirst) { }
+#endif
+
 			size_type      size()  const EA_NOEXCEPT { return mArraySize; }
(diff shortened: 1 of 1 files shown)


include/EASTL/initializer_list.h
```diff
@@ -56,6 +56,14 @@
 			initializer_list() EA_NOEXCEPT  // EA_NOEXCEPT requires a recent version of EABase.  
 			  : mpArray(NULL), mArraySize(0) { }
 
+#if defined(EA_COMPILER_MSVC)
+			// MSVC generates constructor calls with two pointers instead of one pointer + size. The constructor is
+			// public.
+			// See: https://docs.microsoft.com/en-us/cpp/standard-library/initializer-list-class#initializer_list
+			initializer_list(const_iterator pFirst, const_iterator pLast) EA_NOEXCEPT
```
grojo-ea, line 63: Could/should this constructor be private like the one taking an iterator and size?

mtnpke: No, it does not work. It is also public in MS STL: https://github.com/microsoft/STL/blob/2261f7edb760eb3fe0726187c818b796dc7ea798/stl/inc/initializer_list#L34

`1>C:\_GIT\.....(126,2): error C2248: 'std::initializer_list<eastl::pair<const Key,T>>::initializer_list': cannot access private member declared in class 'std::initializer_list<eastl::pair<const Key,T>>'

jhopkins-ea:
Just curious, are you compiling without any std library implementation at all? `EA_HAVE_CPP11_INITIALIZER_LIST` is 1 by default.

mtnpke:
Yes. I am compiling in kernel mode without any STL implementation (besides EASTL).

mtnpke:
Just wondering, is there anything else holding up the merge of this PR? I believe every question was addressed.

include/EASTL/initializer_list.h
grojo-ea: Thanks for checking!

grojo-ea:
Apologies, it seems this fell off our radar. Thanks for answering our questions, this is OK to merge.

PR #504, opened 2023-01-30
Update EABase submodule

mtnpke:
EABase has had some changes that are relevant for EASTL, such as the addition of EA_PLATFORM_WINDOWS_KERNEL. Update the submodule so that those can actually be used.


grojo-ea:
Apologies for the late reply and thanks for fixing this up.

PR #509, opened 2023-06-07
Declare C++14 as required for building

uilianries:
The EASTL uses C++11 and C++14 features like `static_assert` and `constexpr`, however, there is no assigned standard in CMakeLists.txt, which breaks older compiler like GCC 5. Plus, it provides an non predictable scenario, as may depend on the standard supported by the compiler. 

It would be much safer by adding the minimal C++ standard required to build EASTL, so people can reproduce their scenario when building EASTL.


grojo-ea:
Hello,

This looks good to me. 

I noticed we don't have a CLA for you, would you mind [filling out the form](https://electronicarts.na1.echosign.com/public/esignWidget?wid=CBFCIBAA3AAABLblqZhByHRvZqmltGtliuExmuV-WNzlaJGPhbSRg2ufuPsM3P0QmILZjLpkGslg24-UJtek*)? Then we can merge this when the conflicts are resolved.

Thanks

uilianries:
@grojo-ea It's done, the CLA is signed and conflicts are solved. Thank you.

grojo-ea:
Thanks!

Issue #511, opened 2023-06-16
Inquiry about the expected behavior of the comparison operators in RingBuffer


ChaiByte:
[Current code](https://github.com/electronicarts/EASTL/blob/c6141b94d763d2f536bddf40d6abaf878f7ab7ac/include/EASTL/bonus/ring_buffer.h#LL1505C1-L1521C3)

```cpp

template <typename T, typename Container, typename Allocator>
inline bool operator==(const ring_buffer<T, Container, Allocator>& a, const ring_buffer<T, Container, Allocator>& b)
{
return (a.size() == b.size()) && (a.c == b.c);
}
```

I have been using your RingBuffer library and recently came across the implementation of the comparison operators, specifically the operator== and operator<=>. However, I noticed that these operators rely on the behavior of the underlying container's comparison operations, rather than considering the circular nature of the RingBuffer.

Based on my understanding, the comparison operators should take into account the circular behavior of the RingBuffer, as the order and position of elements are crucial for accurate comparison. I believe that the current implementation does not reflect this behavior accurately.

To address this concern, I have made some modifications to the code, which involve comparing the elements of the RingBuffer using iterators and considering the circular characteristics. Here are the modified implementations:

```cpp
template <typename T, typename Container, typename Allocator>
inline bool operator==(const ring_buffer<T, Container, Allocator>& a, const ring_buffer<T, Container, Allocator>& b) {
  return eastl::equal(a.begin(), a.end(), b.begin(), b.end());
}
```

I would like to kindly request your input regarding the expected behavior of the comparison operators in the RingBuffer library. Could you please clarify the intended behavior for these operators, specifically considering the circular nature of the RingBuffer?

---

Test Case:

```cpp
  ring_buffer<int> rb(3);
  rb.assign({0, 1, 2, 3, 4});
  EXPECT_EQ(rb.size(), 3);
  EXPECT_EQ(rb.capacity(), 3);
  for (int i = 0; i < 3; ++i) {
    EXPECT_EQ(rb[i], i + 2);
  }
  
  ring_buffer<int> rb2({3, 4, 2});
  EXPECT_EQ(rb.size(), 3);
  EXPECT_EQ(rb.capacity(), 3);

  ring_buffer<int> rb3({2, 3, 4});
  EXPECT_EQ(rb.size(), 3);
  EXPECT_EQ(rb.capacity(), 3);
  
  EXPECT_TRUE(rb == rb2);   // confused
[...]

PR #514, opened 2023-07-06
Fix ring_buffer comparation logic

ChaiByte:
Make sure to compare two ring buffers in ring-buffer-order.

Resolve: https://github.com/electronicarts/EASTL/issues/511 
CC: @grojo-ea @jhopkins-ea

### DIFF
--- a/include/EASTL/bonus/ring_buffer.h
+++ b/include/EASTL/bonus/ring_buffer.h
@@ -1505,19 +1505,14 @@ namespace eastl
 	template <typename T, typename Container, typename Allocator>
 	inline bool operator==(const ring_buffer<T, Container, Allocator>& a, const ring_buffer<T, Container, Allocator>& b)
 	{
-		return (a.size() == b.size()) && (a.c == b.c);
+		return (a.size() == b.size()) && eastl::equal(a.begin(), a.end(), b.begin());
 	}
 
 
 	template <typename T, typename Container, typename Allocator>
 	inline bool operator<(const ring_buffer<T, Container, Allocator>& a, const ring_buffer<T, Container, Allocator>& b)
 	{
-		const typename ring_buffer<T, Container, Allocator>::size_type sizeA = a.size();
-		const typename ring_buffer<T, Container, Allocator>::size_type sizeB = b.size();
-
-		if(sizeA == sizeB)
-			return (a.c < b.c);
-		return sizeA < sizeB;
+		return eastl::lexicographical_compare(a.begin(), a.end(), b.begin(), b.end());
 	}
 
 
--- a/test/source/TestRingBuffer.cpp
+++ b/test/source/TestRingBuffer.cpp
@@ -549,6 +549,56 @@ int TestRingBuffer()
 		EATEST_VERIFY(rbVectorInt[5] == 4);
 	}
 
+	{
+		// Comparation operator ==, operator <
+		// Fix bug mentioned in https://github.com/electronicarts/EASTL/issues/511
+		typedef ring_buffer<int, vector<int>> RBVectorInt;
+
+		RBVectorInt rbVectorInt(3);
+
+		rbVectorInt.push_back(0);
+		rbVectorInt.push_back(1);
+		rbVectorInt.push_back(2);
+		rbVectorInt.push_back(3);
+		rbVectorInt.push_back(4);
+
+		EATEST_VERIFY(rbVectorInt[0] == 2);
+		EATEST_VERIFY(rbVectorInt[1] == 3);
+		EATEST_VERIFY(rbVectorInt[2] == 4);
+
+		RBVectorInt rbVectorInt2({3, 4, 2});
+		RBVectorInt rbVectorInt3({2, 3, 4});
+		RBVectorInt rbVectorInt4({3, 4, 5});
+
+		EATEST_VERIFY(rbVectorInt != rbVectorInt2);
+		EATEST_VERIFY(rbVectorInt == rbVectorInt3);
+		EATEST_VERIFY(rbVectorInt < rbVectorInt4);
+
+		// Different size
+		RBVectorInt rbVectorInt5({1, 2});
+		EATEST_VERIFY(rbVectorInt != rbVectorInt5);
+		EATEST_VERIFY(rbVectorInt > rbVectorInt5);
+
+		RBVectorInt rbVectorInt6({2, 3});
+		EATEST_VERIFY(rbVectorInt != rbVectorInt6);
+		EATEST_VERIFY(rbVectorInt > rbVectorInt6);
+
+		RBVectorInt rbVectorInt7({3, 4});
+		EATEST_VERIFY(rbVectorInt != rbVectorInt7);
+		EATEST_VERIFY(rbVectorInt < rbVectorInt7);
+
+		RBVectorInt rbVectorInt8({1, 2, 3, 4});
+		EATEST_VERIFY(rbVectorInt != rbVectorInt8);
+		EATEST_VERIFY(rbVectorInt > rbVectorInt8);
+
+		RBVectorInt rbVectorInt9({2, 3, 4, 5});
+		EATEST_VERIFY(rbVectorInt != rbVectorInt9);
+		EATEST_VERIFY(rbVectorInt < rbVectorInt9);
+
+		RBVectorInt rbVectorInt10({3, 4, 5, 6});
+		EATEST_VERIFY(rbVectorInt != rbVectorInt10);
+		EATEST_VERIFY(rbVectorInt < rbVectorInt10);
+	}
 
 	{
 		EA::UnitTest::Rand rng(EA::UnitTest::GetRandSeed());


include/EASTL/bonus/ring_buffer.h
```diff
@@ -1505,7 +1505,12 @@ namespace eastl
 	template <typename T, typename Container, typename Allocator>
 	inline bool operator==(const ring_buffer<T, Container, Allocator>& a, const ring_buffer<T, Container, Allocator>& b)
 	{
-		return (a.size() == b.size()) && (a.c == b.c);
+		const typename ring_buffer<T, Container, Allocator>::size_type sizeA = a.size();
+		const typename ring_buffer<T, Container, Allocator>::size_type sizeB = b.size();
+
+		if(sizeA == sizeB)
+			return eastl::equal(a.begin(), a.end(), b.begin());
+		return sizeA < sizeB;
```
grojo-ea, line 1513: I don't think this is correct. Shouldn't this unconditionally return `false` if the ring buffers have different sizes?

test/source/TestRingBuffer.cpp
```diff
@@ -549,6 +549,31 @@ int TestRingBuffer()
...
+		EATEST_VERIFY(rbVectorInt[1] == 3);
+		EATEST_VERIFY(rbVectorInt[2] == 4);
+
+		RBVectorInt rbVectorInt2({3, 4, 2});
+		RBVectorInt rbVectorInt3({2, 3, 4});
+		RBVectorInt rbVectorInt4({3, 4, 5});
+
+		EATEST_VERIFY(rbVectorInt != rbVectorInt2);
+		EATEST_VERIFY(rbVectorInt == rbVectorInt3);
+		EATEST_VERIFY(rbVectorInt < rbVectorInt4);
```
grojo-ea, line 575: Could you add some tests with ring buffers of different sizes? Both for equality and comparison. And add tests for comparing in the opposite direction as well.

include/EASTL/bonus/ring_buffer.h
```diff
@@ -1516,7 +1521,7 @@ namespace eastl
 		const typename ring_buffer<T, Container, Allocator>::size_type sizeB = b.size();
 
 		if(sizeA == sizeB)
-			return (a.c < b.c);
+			return eastl::lexicographical_compare(a.begin(), a.end(), b.begin(), b.end());
 		return sizeA < sizeB;
```
grojo-ea, line 1525: `lexicographical_compare` should handle different size ranges, I think for consistency with other containers (like vector), this should just return the same as the lexicographical comparison. So a ring buffer with `{ 1, 2 }` in it compares smaller to one with `{ 3 }` in it (as it would for vectors, for example)

include/EASTL/bonus/ring_buffer.h
```diff
@@ -1505,7 +1505,12 @@ namespace eastl
 	template <typename T, typename Container, typename Allocator>
 	inline bool operator==(const ring_buffer<T, Container, Allocator>& a, const ring_buffer<T, Container, Allocator>& b)
 	{
-		return (a.size() == b.size()) && (a.c == b.c);
+		const typename ring_buffer<T, Container, Allocator>::size_type sizeA = a.size();
```
grojo-ea, line 1508: This is a minor thing but given that I'm requesting other changes. This is probably a good place to use `const auto` .

grojo-ea (changes requested):

include/EASTL/bonus/ring_buffer.h
ChaiByte: You are right，it is an obvious mistake. 😅 Fixed.

I have found that it may be possible to avoid calling `size()` here and instead use the `eastl::identical` interface for more precise semantics and readability. However, it seems that this interface is not mandated by the standard. Which approach do you think is better?

```cpp
// Method A
return eastl::identical(a.begin(), a.end(), b.begin(), b.end());

// Method B
return (a.size() == b.size()) && eastl::equal(a.begin(), a.end(), b.begin(), b.end());

// Method C 
if (a.size()!= b.size()
    return false;
return eastl::equal(a.begin(), a.end(), b.begin(), b.end());
```

And it seems redundant to separately getting size variables in current implement:

```cpp
const auto sizeA = a.size();
const auto sizeB = b.size();
```

They are just used once.

include/EASTL/bonus/ring_buffer.h
ChaiByte: as previously mentioned

include/EASTL/bonus/ring_buffer.h
ChaiByte: Done.

test/source/TestRingBuffer.cpp
ChaiByte: Sure. To keep test minimal and simple, I just add one more var `rbVectorInt5` which has different size.

include/EASTL/bonus/ring_buffer.h
jhopkins-ea: `eastl::identical` is equivalent to `std::equal(first1, last1, first2, last2)`, ie. overloads 5-8 listed at cppref [here](https://en.cppreference.com/w/cpp/algorithm/equal). Method A and B have the same complexity from a Big O notation perspective. Please use method A.

ChaiByte: @jhopkins-ea Thank you for sharing this information with me.

I agree that method B and C have the same Big O complexity. The difference between them lies solely in the number of times the instructions are executed. and I trust the conditional branch should eventually be optimized away by the compiler. 

Let's talking more about the method A, https://github.com/electronicarts/EASTL/blob/61c6a018e87d2fa1fde43f1ede813f15e8758dfb/include/EASTL/algorithm.h#L1960-L1969

Assume that here are two ring buffer `rb1` (with size 1'000'000) and `rb2` (with size 1'000'010):

```cpp
RBVectorInt rb1 {0, ..., 0} // all elements are 0
RBVectorInt rb2 {0, ..., 0, 1, 0, 0, ..., 0,} // the 1'000'001th element is 1, others are 0
```

In this case, `eastl::identical` need at least 1'000'000 times comparation to get result. But if we can get size information in advance, we can avoid redundant comparation.

ChaiByte: Let's see what cppref says in [equal](https://en.cppreference.com/w/cpp/algorithm/equal):

> Complexity
> 5,7) At most min(last1 - first1, last2 - first2) applications of the predicate.
However, if InputIt1 and InputIt2 meet the requirements of [LegacyRandomAccessIterator](https://en.cppreference.com/w/cpp/named_req/RandomAccessIterator) and last1 - first1 != last2 - first2 then no applications of the predicate are made (size mismatch is detected without looking at any elements).

See GNU libstdc++ [code](https://github.com/gcc-mirror/gcc/blob/c496d15954cdeab7f9039328f94a6f62cf893d5f/libstdc%2B%2B-v3/include/bits/stl_algobase.h#L1184), they use a `_BoolType` to check whether inputs are RandomAccessIterator or not to optimize specific case.

I guess that's why they don't implement a separate method called `identical`. The interface `identical` currently provided will allow me to form a good habit of comparing the size before calling.

grojo-ea: > eastl::identical is equivalent to std::equal(first1, last1, first2, last2), ie. overloads 5-8 listed at cppref [here](https://en.cppreference.com/w/cpp/algorithm/equal).

I just realized EASTL doesn't have the `equal` overloads taking 2 iterator pairs. Also, given that `eastl::identical` doesn't do the range size optimization for random access iterators it's technically not the same.

I think I'd prefer us to add those overloads (with the correct optimizations for random access iterators) than to proliferate the usage of `eastl::identical` (which arguably should be deprecated in favor of the new `equal` overloads). But this is out of the scope of this PR.

Given the current state of things, option B is probably the best we can do.

test/source/TestRingBuffer.cpp
grojo-ea: I disagree that tests should be minimal and simple. Tests should be as complicated as they need to be ;).

This new test is good but we're still missing a test where one container is a suffix of the other. so comparing `rbVectorInt` against `{2, 3}` or `{2, 3, 4, 5}` for example, and a test where the size is smaller but the lexicographical order is greater, like against `{3, 2}`.

include/EASTL/bonus/ring_buffer.h
ChaiByte: > I just realized EASTL doesn't have the `equal` overloads taking 2 iterator pairs. Also, given that `eastl::identical` doesn't do the range size optimization for random access iterators it's technically not the same.

I can provide more information:

When I first find the `eastl::identical` interface, I didn't directly realize that it is an overload of the `equal` function that accepts different parameters. I subconsciously think that this is a "more efficient" `equal`, and the user must ensure that the length is equal before calling. for better performance.

Thanks to @jhopkins-ea for reminding me to establish right perception. XD


> I think I'd prefer us to add those overloads (with the correct optimizations for random access iterators) than to proliferate the usage of `eastl::identical` (which arguably should be deprecated in favor of the new `equal` overloads). But this is out of the scope of this PR.

It would be great if this could be done, I tried to write `eastl::equal(a.begin(), a.end(), b.begin(), b.end())` at first but found that it could not be compiled , that feels strange, because it can be done in other STL implementations, and it will be a better experience to be consistent.

> Given the current state of things, option B is probably the best we can do.

Glad to hear that, are there other suggestions? I think it's time for merging.

test/source/TestRingBuffer.cpp
ChaiByte: Aye Captain, additional necessary test cases have been equipped.

grojo-ea (approved):
Thanks for the changes, the contribution, and the discussion.

include/EASTL/bonus/ring_buffer.h
jhopkins-ea: Good pick up on the complexity differences between `identical` and `equal`. We should in the future deprecate identical in favour of new equal overloads that have the complexity guarantees.

This is a minor issue, but `container.size()`/`std::size()` does not have a guaranteed constant time, as opposed to `ranges::size()` ([see here](https://en.cppreference.com/w/cpp/ranges/size)). If we used a `eastl::list` as the backing container this would be a linear time operation, because `eastl::list` does not store the size (unlike a `std::list`).

Thanks for your contribution.

PR #515, opened 2023-07-07
Reduce complexity when inputer iterator meet RandomAccessIterator

ChaiByte:
As described in https://github.com/electronicarts/EASTL/pull/514#discussion_r1255556282 , It's a X-Y problem when discussing the correct method fixing ring buffer comparation operator. Another issue to become evident.

Current `eastl::identical` is not effient enough when input iterators meet [LegacyRandomAccessIterator](https://en.cppreference.com/w/cpp/named_req/RandomAccessIterator) requirements.

Refs: https://en.cppreference.com/w/cpp/algorithm/equal

> Complexity
> 5,7) At most min(last1 - first1, last2 - first2) applications of the predicate.
> However, if InputIt1 and InputIt2 meet the requirements of [LegacyRandomAccessIterator]> (https://en.cppreference.com/w/cpp/named_req/RandomAccessIterator) and last1 - first1 != last2 - first2 then no applications of the predicate are made (size mismatch is detected without looking at any elements).
> 2,4,6,8) same, but the complexity is specified as O(x), rather than "at most x".

By the way, I want to know why did EASTL choose to implement the `equal` and `identical` interfaces separately at the beginning of the design, is there any reference to explain the reason?

### DIFF
--- a/include/EASTL/algorithm.h
+++ b/include/EASTL/algorithm.h
@@ -1960,6 +1960,10 @@ namespace eastl
 	bool identical(InputIterator1 first1, InputIterator1 last1,
 				   InputIterator2 first2, InputIterator2 last2)
 	{
+		auto d1 = eastl::distance(first1, last1);
+		auto d2 = eastl::distance(first2, last2);
+		if(d1 != d2)
+			return false;
 		while((first1 != last1) && (first2 != last2) && (*first1 == *first2))
 		{
 			++first1;
@@ -1975,6 +1979,10 @@ namespace eastl
 	bool identical(InputIterator1 first1, InputIterator1 last1,
 				   InputIterator2 first2, InputIterator2 last2, BinaryPredicate predicate)
 	{
+		auto d1 = eastl::distance(first1, last1);
+		auto d2 = eastl::distance(first2, last2);
+		if(d1 != d2)
+			return false;
 		while((first1 != last1) && (first2 != last2) && predicate(*first1, *first2))
 		{
 			++first1;


include/EASTL/algorithm.h
```diff
@@ -1960,6 +1960,10 @@ namespace eastl
 	bool identical(InputIterator1 first1, InputIterator1 last1,
 				   InputIterator2 first2, InputIterator2 last2)
 	{
+		auto d1 = eastl::distance(first1, last1);
```
grojo-ea, line 1963: We shouldn't be calling `distance` unconditionally here for at least these 2 reasons:
- `distance` is linear in complexity for iterators which are not random access. For example, if these were iterators over a singly liked list, this will iterate both lists fully to compute the distances, which is not necessary for an equality check if for example the first element of the lists differ.
- input iterators are not guaranteed to be multi-pass. In other words, the call to `distance` might "consume" the input iterators so that we can't use them again to iterate for the comparisons below.

grojo-ea:
I agree that there's room for optimizations here in the case of random access iterators, but this implementation is not the right one for iterators which are not random access. 

This optimization needs to branch behaviour when the iterator is random access vs when it isn't. EASTL doesn't yet require C++17 to compile so at this point this should probably be done with tagged dispatch (similar to what we do with `eastl::distance`) or with a "normal-not-constexpr" if on a compile time constant check,  as opposed to using `if constexpr`, which would be the way to do this in C++17.

include/EASTL/algorithm.h
ChaiByte: You are right, ideally we should try to use iterator_traits to get tags and provide at least two versions of the implementation, the idea in my head is:

```cpp
	template <typename InputIterator>
	EA_CONSTEXPR
	bool identical_impl(InputIterator first1, InputIterator last1,
						InputIterator first2, InputIterator last2,
	 					EASTL_ITC_NS::input_iterator_tag)
	{
		while((first1 != last1) && (first2 != last2) && (*first1 == *first2))
		{
			++first1;
			++first2;
		}
		return (first1 == last1) && (first2 == last2);
	}

	template <typename RandomAccessIterator>
	EA_CONSTEXPR
	bool identical_impl(RandomAccessIterator first1, RandomAccessIterator last1,
						RandomAccessIterator first2, RandomAccessIterator last2,
						EASTL_ITC_NS::random_access_iterator_tag)
	{
		auto d1 = eastl::distance(first1, last1);
		auto d2 = eastl::distance(first2, last2);
		if(d1 != d2)
			return false;
		while((first1 != last1) && (first2 != last2) && (*first1 == *first2))
		{
			++first1;
			++first2;
		}
		return (first1 == last1) && (first2 == last2);
	}

	template <typename InputIterator1, typename InputIterator2>
	EA_CONSTEXPR
	bool identical(InputIterator1 first1, InputIterator1 last1,
				   InputIterator2 first2, InputIterator2 last2)
	{
		typedef typename eastl::iterator_traits<InputIterator1>::iterator_category IC1;
		typedef typename eastl::iterator_traits<InputIterator2>::iterator_category IC2;
		typedef typename eastl::common_type<IC1, IC2>::type IC;
		return identical_impl(first1, last1, first2, last2, IC());
	}


	/// identical
	///
	template <typename InputIterator, typename BinaryPredicate>
	EA_CONSTEXPR
	bool identical_impl(InputIterator first1, InputIterator last1,
						InputIterator first2, InputIterator last2,
	 					BinaryPredicate predicate, EASTL_ITC_NS::input_iterator_tag)
	{
		while((first1 != last1) && (first2 != last2) && predicate(*first1, *first2))
		{
			++first1;
			++first2;
		}
		return (first1 == last1) && (first2 == last2);
	}

	template <typename RandomAccessIterator, typename BinaryPredicate>
	EA_CONSTEXPR
	bool identical_impl(RandomAccessIterator first1, RandomAccessIterator last1,
						RandomAccessIterator first2, RandomAccessIterator last2,
						BinaryPredicate predicate, EASTL_ITC_NS::random_access_iterator_tag)
	{
		auto d1 = eastl::distance(first1, last1);
		auto d2 = eastl::distance(first2, last2);
		if(d1 != d2)
			return false;
		while((first1 != last1) && (first2 != last2) && predicate(*first1, *first2))
		{
			++first1;
			++first2;
		}
		return (first1 == last1) && (first2 == last2);
	}

	template <typename InputIterator1, typename InputIterator2, typename BinaryPredicate>
	EA_CONSTEXPR
	bool identical(InputIterator1 first1, InputIterator1 last1,
				   InputIterator2 first2, InputIterator2 last2, BinaryPredicate predicate)
	{
		typedef typename eastl::iterator_traits<InputIterator1>::iterator_category IC1;
		typedef typename eastl::iterator_traits<InputIterator2>::iterator_category IC2;
		typedef typename eastl::common_type<IC1, IC2>::type IC;
		return identical_impl(first1, last1, first2, last2, predicate, IC());
	}
```

Currently I may not have enough time to polish and discuss the implementation details.
I wull focus on #514 for now, and then participate in the discussion on `equal` and `identical` later.

ChaiByte:
I'll close this PR temporarily until we find a better solution.

PR #526, opened 2024-03-26
Create dependabot.yml

IsakTheHacker:
I saw that some dependencies in the GitHub action workflows were out of date. Since GitHub is always migrating to newer Node runtimes it is important to keep GitHub action dependencies up to date so they won't become incompatible all of a sudden.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,7 @@
+version: 2
+updates:
+- package-ecosystem: "github-actions"
+  directory: "/"
+  schedule:
+    interval: "weekly"
+  open-pull-requests-limit: 5


grojo-ea:
Hello,

Thanks for the PR, if I'm getting [this] (https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/about-dependabot-version-updates#github-actions) right, this relates to the fact that we're using `actions/checkout@v2`, `actions/upload-artifact@v2.3.1`, and `actions/download-artifact@v2.1.0` in our `c-cpp.yml` file but there's already newer versions of those actions that we should be using instead, and your change would make it so that we get notified when this happens?

If so, that's neat!

I noticed we don't yet have a [CLA form](https://electronicarts.na1.echosign.com/public/esignWidget?wid=CBFCIBAA3AAABLblqZhByHRvZqmltGtliuExmuV-WNzlaJGPhbSRg2ufuPsM3P0QmILZjLpkGslg24-UJtek*) for you, we'd need that in our records before we can merge this.

PR #533, opened 2024-06-24
array: add bounds checks for operator[]

tempname11:
Many other containers in this library (including vector) do have assertions for operator[], and so it makes sense to have them in array as well.


jhopkins-ea:
Hi. We added this recently to our internal version and it will be included in the next public release. No plans as to when that release is yet.

Issue #506, opened 2023-02-14
No signed tarball for releases


jesper-arista:
Hello,

There is no signed files for your releases, any chance you could sign the latest release, it would help using this code in other 3rd party packages.

thanks
Jesper

PR #536, opened 2024-08-16
EASTL version 3.21.23

grojo-ea:
Updates since the last version.
 * removed generic definition of `numeric_limits<>`
 * removed EASTL_CUSTOM_FLOAT_CONSTANTS_REQUIRED & associated functionality
 * removed `Internal::numeric_limits_base`
 * added `isnan` to EASTL when compiled with C++20
 * added `is_constant_evaluated`
 * fixed `midpoint` to behave correctly when passed NaNs
 * Add bounds checking assert for `array<T, N>::operator[]`.
 * Fixed `fixed_vector` move constructor with an explicit allocator parameter. It now works with move-only types.
 * Add `bit_and`, `bit_or`, `bit_xor`, `bit_not` and `owner_less<void>` function objects.
 * Use string builtins to improve codegen for string literals operations on `string_view` and `string`.
 * `span` is now specialized for static extents so that it doesn't need to store the size at runtime.
 * Fixed a bug where calling `emplace`, `emplace_hint`, or `try_emplace` on a `string_hash_map` would call the underlying `hash_map` implementation, which caused a crash when `string_hash_map` would try to free memory it didn't own.
 * Use EA_DEPRECATED date based macros if available to guide deprecations.
 * Consolidate implementations of `CharTypeStringSearch/Find` functions in `char_traits.h`.
 * Minor optimization to `CharTypeStringRSearch` to reduce iteration loop count.
 * Fix `deque` move assignment operator for move only and non-default constructible types.
 * Added monadic operations `and_then`, `transform` and `or_else` for `optional`.
 * Fixed `value_or` for r-value refs to `optional`.
 * Various `bitset` fixes and improvements:
   * Replace `bitset(uint32_t)` with `bitset(unsigned long long)` which matches the standard and means that 64 bit integers can be correctly converted to a bitset. As a result `from_uint32()` and `from_uint64()` are now unnecessary (not deprecated).
   * Add `to_ulong_checked()` with the same behaviour as the standard, ie. throws or asserts if the target type cannot represent all the set bits.
   * Add an extension to the standard, `as_uint<T>()` / `as_ulong()`, cast to a unsigned integral that can represent the entire bitset. If the target type cannot represent the entire bitset, then issue a compile error (overload does not exist).
   * Fix implementation for word types smaller than 32 bit. Fixed logic in find_prev(), other iteration functions and conversion to/from bitset functions that did not work for such types.
   * No longer assume `sizeof(word_type) == EA_PLATFORM_WORD_SIZE`. eg. `from_uint64()` previously would not copy the entire value into a `bitset<N, uint32_t>` for `N > 32`.
   * Now compiles for a word type of `unsigned long` or `unsigned long long` where the type is not `uint64_t`.
   * Add an assert that the word type is an unsigned integral, as these are the only types that are supported and tested.
 * Add `get<T>(pair<T1, T2>)`. Make `get<I>(pair<T1, T2>)` constexpr.
 * Fix variant non-member functions `get_if()`, `get()`, `holds_alternative()` and converting constructor and assignment so that they only conditionally compile when the index or type unambiguously refers to a variant alternative.
 * `any::emplace()` now returns a reference.
 * `lru_cache::emplace()` now compiles correctly, forwarding its arguments and returns an pair of iterator and bool indiciating emplace success.
 * Made changes to avoid `-Wdeprecated-copy-with-user-provided-copy`, which manifests for the following reasons:
    * definition of implicit copy assignment operator for `class` is deprecated because it has a user-provided copy constructor
    * definition of implicit copy constructor for `class` is deprecated because it has a user-provided copy assignment operator
 * Fix various issues with `segmented_vector`:
    * Fix copy/move construction/assignment so it doesn't do a shallow copy.
    * `clear()` now behaves like other `eastl` containers in that it does not free memory when called, the container maintains it's capacity.
    * `segmented_vector` now supports types which are not default constructible.
    * Added some missing API functions: `reserve`, `resize`, `shrink_to_fit`, `emplace_back`.
    * Added comparison operations.
  * Add constructor overloads to `map`, `set`, `hash_set`, `hash_map` that take an `initializer_list` and Allocator.
  * Enable some container asserts (when EASTL_ASSERT_ENABLED is enabled), removing the requirement for EASTL_EMPTY_REFERENCE_ASSERT_ENABLED. When called with invalid arguments these functions will result in undefined behaviour.
  * Add some additional asserts to container adaptors.
 * Fix undefined behaviour in `shell_sort()` algorithm.
 * Add return values for `deque`, `list`, `slists`'s member functions `emplace_front()`, `emplace_back()` and `unique()`.
 * Add `slist::unique()` function.
 * Deprecate `make_pair_ref()`, use `make_pair()` instead. Also deprecate a `make_pair()` overload which was only enabled on MSVC.
 * Deprecate `stack::emplace_back()`, `use stack::emplace()` instead.
 * Fix references to std library that should be eastl.
 * Fix compilation issues when C4626 is enabled as error on MSVC.
 * Deprecate `vector_map<Key, T>::at(index)`. this function was accidentally inherited from `vector` with incorrect semantics.
 * Optimize bitset's `find_first()`, `find_next()`, `find_last()`, `find_prev()` (which are extensions to the standard).
 * Fix `bitset<N, eastl_uint128_t>::find_last()`.
 * Added `erase_unordered` and `erase_unordered_if` for `vector` and `deque` (`fixed_vector` works implicitly with the
   `vector` version since it is a `vector` with a special allocator and these functions don't affect capacity). These
   work similar to free function `erase`/`erase_if` but are more efficient if those would result in moving a lot of
   elements. They achieve this by not respecting the order of the remaining element and just moving the elements from
   the end into the now empty spots.
 * `optional::emplace` now correctly returns `T&` instead of `void` as specified in the standard.
 * Explicitly define special member functions for some types. This fixes new compile errors from Visual Studio 17.7.0 Preview.
 * Deprecate uninitialized_copy_ptr et al - uninitialized_copy is already perfectly suitable for pointers.
 * Deprecate unwrap_iterator, is_reverse_iterator, is_move_iterator, is_insert_iterator, generic_iterator (the last is internal and should already be unused).
 * Fix `function::target` not compiling due to missing template parameter when we called into the base class' member function.


vikhik:
Can this be `git tag`'d with the new version number, please, so we can request new package updates on vcpkg/conan?

Also, there's an outstanding issue to sha sign releases that might be nice to tackle: https://github.com/electronicarts/EASTL/issues/506

vikhik:
Can this be `git tag`'d with the new version number, please, so we can request new package updates on vcpkg/conan?

Also, there's an outstanding issue to sha sign releases that might be nice to tackle: https://github.com/electronicarts/EASTL/issues/506

@jhopkins-ea 👋 say hi to the team for me 😁

xq114:
Could you also make a release tag in EABase https://github.com/electronicarts/EABase so this package can rely on correct version of EABase?

PR #537, opened 2024-08-20
Fix/allocator

prikolium-cfx:
This PR implementing new simple custom allocator that using _aligned_malloc on Windows and memalign on linux. Should fix issue with performance and deallocation on Windows.


prikolium-cfx:
Oops, accidently opened PR

Issue #539, opened 2024-09-01
Please consider to use git submodules instead of FetchContent


yurivict:
File fetching isn't allowed during package build in any sane packaging system.
For this reason FetchContent isn't compatible with packaging.

Please consider going back to git submodules.

grojo-ea:
Hello,

Thanks for bringing this up. We were trying the FetchContent approach to try to break the circular submodule dependency issue when cloning EASTL with `--recursive` (see https://github.com/electronicarts/EASTL/issues/410 and https://github.com/electronicarts/EASTL/issues/433). I was unaware that this would break packaging. 

I'm unfamiliar with how package managers do the packaging process. Would it be enough to make it so that we can build/use EASTL itself without FetchContent but keep the FetchContent declarations in  for EASTLTest?

If that's the case I think we can potentially make EABase use FetchContent for its testing project and bring in EABase as a submodule. This is the only dependency that's needed to build EASTL and the test projects (which are the ones that generate the submodule cycles) could still use FetchContent.

yurivict:
I maintain the FreeBSD port for EASTL and in the ports framework the cmake setup is such that FetchContent just doesn't work, even for tests.
FetchContent instructions need to be patched away and replaced with pre-downloaded directories.

Maybe it would be a good idea to keep both git submodules and FetchContent? FetchContent can be activated when the directories aren't present.

Before the last update git submodules worked well for the port, regardless of any possible circular depencies.
Circular dependencies are essentially a bug in the repositories that have them. These problems should just be fixed (circular dependencies broken)?

grojo-ea:
> the cmake setup is such that FetchContent just doesn't work, even for tests.

I guess a more specific question is: are the tests build and run as part of the packaging process? What I'm suggesting with using FetchContent only in test projects wouldn't invoke FetchContent unless you invoke CMake explicitly with `-DEASTL_BUILD_TESTS=ON` (or its equivalent in other libraries)

> Circular dependencies are essentially a bug in the repositories that have them. These problems should just be fixed (circular dependencies broken)?

Agreed, that's what we're trying to fix. For some context we don't use CMake nor submodules internally, our test projects  use the various libraries liberally but none of the tests projects depend on each other and no library depends on any test projects so there aren't really any circular dependencies, the problem is that we want the library an its corresponding tests in the same repository, and if we add the test's dependencies as submodules in the libray's repository that's where the circular dependencies arise.

yurivict:
[...]

Issue #410, opened 2021-01-03
Submodule cloning error.


CybernetHacker14:
First of all, the effort to establish such a great piece of technology and then to open source it, is a great gesture, much appreciated.

I tried adding this repository as a submodule to an existing project. It got cloned properly. However, the nested submodules inside this repository - EABase, EAStdC, didn't got cloned properly - there was an error during updating the submodules of EASTL.

The messages stated things as - 
```
please make sure you have the correct access rights and the repository exists.
```  
and
```
fatal: clone of 'git@github.com:electronicarts/EAAssert.git' into submodule path '<path-to-EASTL>/test/packages/EAThread/test/packages/EAAssert' failed
``` 
turned up. This was for all submodules of EASTL. 

Does adding EASTL as a submodule require operating the repository through `SSH` only? Is `HTTPS` not supported?

ChuckStarchaser:
I've been having all kinds of problems similar to this, and wasn't using git, or installing as a module.  First I tried to install it using conan, and there were no cpp files anywhere, and no CMakeList.txt files. Could not find a solution.
Then I tried downloading the zip file, but the 6 sub-packages under test/packages are empty folders in the zip.
I was about to try git, but came to look at the issues first, and found yours.

ChuckStarchaser:
UPDATE:  I managed to solve my problem.  I discovered that each of those 6 subpackages, if you go into the folder in github the green Code button lights up and you can download as zip or git the particular library.  Once I did, CMake worked, then Make worked.

CybernetHacker14:
I was trying to add this repo as a submodule, and this repo also has submodules inside it. If someone wants to checkout the latest updates to whatever was done, he/she has to go to great lengths to have all of the submodules checked out on latest version.

zkxjzmswkwl:
This is still fucked.

iyadahmed:
> This is still fucked.

yes

iyadahmed:
> UPDATE: I managed to solve my problem. I discovered that each of those 6 subpackages, if you go into the folder in github the green Code button lights up and you can download as zip or git the particular library. Once I did, CMake worked, then Make worked.

this seems to be a nice workaround for now

omartijn:
The submodules are really messed up, because they are recursive. There is a circular reference between some of the submodules, so the clone never ends, and it'll keep going deeper and deeper into subdirectories with endless copies of submodule repos.

PR #541, opened 2024-09-09
Bring back submodules, but use the no-test branches.

grojo-ea:
As reported in https://github.com/electronicarts/EASTL/issues/539, CMakes `FetchContent` does not play well with package managers so we're trying going back to submodules for the dependencies.

However, we still want to fix the circular dependency debacle https://github.com/electronicarts/EASTL/issues/410. Given that the circular dependencies are in the test projects, I've created `no-test` branches in each of the dependencies and sync those branches in the submodule dependencies.

### DIFF
--- a/.github/workflows/c-cpp.yml
+++ b/.github/workflows/c-cpp.yml
@@ -8,12 +8,13 @@ on:
 
 jobs:
   checkout:
-    name: Checkout EASTL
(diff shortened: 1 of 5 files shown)


vikhik:
I would vote against this - keeping an up-to-date "no-test" branch is a trap from a git-user perspective.

PR #547, opened 2024-12-03
Protect string comparisons against segfaults

### LINKED ISSUE #546: string.h: Missing nullptr handling in operator== leading to segfaults
Users of the EASTL might compare a string object to nullptr. While this is actually a useless operation, given string objects cannot be NULL, it may occurr in templated code and ripple down to the point where `CharStrlen` from `char_traits.h` is called with a nullptr, triggering a segfault.

A real life example for this is when EASTL is used with the `trompeloeil` mocking framework.

Minimal demo:
```
#include <EASTL/fixed_string.h>

int main() {
    constexpr size_t minSize{2U};
    eastl::fixed_string<char, minSize, false> fs;
    if (fs == nullptr)
    {
        // This can never occur.
    }
}
```

Imho, there shall be a check for null in the `operator==` implementation(s).

JSydll:
This intends to resolve #546. 

Comparing string objects to `nullptr` is not prevented by the implementation. Therefore, it should check before starting the actual comparison - which currently leads to a segfault due to the unprotected use of `CharStrlen`. Given that `CharStrlen` fulfills the standard specification of showing UB when called like that, the issue should be fixed on the calling site.

The first commit introduces minimal test cases for the global equality operators for `basic_string`, the second the (quite small) fix.

### DIFF
--- a/include/EASTL/string.h
+++ b/include/EASTL/string.h
@@ -3542,6 +3542,8 @@ namespace eastl
 	template <typename T, typename Allocator>
 	inline bool operator==(const typename basic_string<T, Allocator>::value_type* p, const basic_string<T, Allocator>& b)
 	{
+		if (p == nullptr)
+			return false;
 		typedef typename basic_string<T, Allocator>::size_type string_size_type;
 		const string_size_type n = (string_size_type)CharStrlen(p);
 		return ((n == b.size()) && (Compare(p, b.data(), (size_t)n) == 0));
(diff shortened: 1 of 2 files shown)


virt00l:
Im pretty sure that's intentionaly for perfomance sake
`strlen("foo", NULL)` would also segfault (I assuming for the same reason)

JSydll:
I do acknowledge that `strlen` (or the EASTL variant of it, `CharStrlen`) may segfault, as this is allowed by the specification. What is far less obvious to me is the described use of comparison operators. The current behavior is especially brittle in template contexts.

Do you have an indication on performance impact in this use case? And maybe a recommendation on an alternative solution?

PR #558, opened 2025-02-04
implement EASTL_[UN]LIKELY on clang (in addition to gcc)

virt00l:
It supported it since forever, there is no point not to use it


virt00l:
@rparolin @grojo-ea @jhopkins-ea Any chance onto reviewing and merging this (and (*)) commits?
(*) PS: https://github.com/electronicarts/EASTL/issues?q=is%3Apr+is%3Aopen+author%3Avirt00l

PR #559, opened 2025-02-04
fix swap of eastl::list<unique_ptr[]>

virt00l:
...or any other type that has deleted dtor.
```
const this_type temp(*this); // << This would attept to call copy ctor of unique_ptr<> which is explicitly deleted
*this = x;
x     = temp;
```


virt00l:
Nvm, this did not fix it. WIP, please don't merge (yet)

PR #571, opened 2025-03-28
fix(clang): Correct literal operator in string_view.h

vikhik:
Fix a new warning in clang (18?) - `-Wdeprecated-literal-operator`

ref: https://github.com/fmtlib/fmt/issues/3607


vikhik:
Should've used a non-master branch. Resubmitting

Issue #542, opened 2024-09-18
Second overload for operator new not documented


vikhik:
Nowadays, EASTL requires two overloaded operator news, however the docs don't specify this.

Please update the integration docs: https://github.com/electronicarts/EASTL/blob/master/doc/CMake/EASTL_Project_Integration.md#overloading-operator-new

PR #508, opened 2023-05-05
Update EASTL_Project_Integration.md

JochenJankowai:
Some info for building with CMake was outdated.

- Adding the subdirectory will set up include paths for the packages. No need to do it explicitly.
- Set include directory per target.
- `EASTL_LIBRARY` is now called `EASTL`.
- Two `new` operator overloads are required.

PR #573, opened 2025-03-29
docs: Add missing overload to the integration guide

vikhik:
Fix for https://github.com/electronicarts/EASTL/issues/542


vikhik:
Done better by https://github.com/electronicarts/EASTL/pull/508

PR #577, opened 2025-06-28
Replaced std::bool_constant with bool_constant

nerdjot:
Replaced std::bool_constant with correct bool_constant.


jhopkins-ea:
fixed in master

PR #578, opened 2025-06-28
Corrected std::integral_constant to eastl::integral_constant in array.h

nerdjot:
Explicitly added eastl namespace to integral_constant to avoid std version.


jhopkins-ea:
fixed in master

